PR #3, opened 2020-12-11
[PF-268] Add github action to publish libraries

yonghaoy:
See Readme: https://github.com/DataBiosphere/terra-common-lib/blob/yyu-PF-github/README.md
for more details. 

Also. the current Migrate package is not the final package we would use. It is just for my testing purpose to verify how dependecy injection works outside Springboot..

Tested it here: https://broadinstitute.jfrog.io/ui/repos/tree/General/libs-snapshot-local%2Fbio%2Fterra%2Fterra-common-lib


.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,80 @@
...
+#          path: build/reports/tests
+      - name: "Bump the tag to a new version"
+        uses: databiosphere/github-actions/actions/bumper@v0.0.3
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DEFAULT_BUMP: patch
+          RELEASE_BRANCHES: master,develop
+          VERSION_FILE_PATH: build.gradle
+          VERSION_LINE_MATCH: ^version
+          VERSION_SUFFIX: SNAPSHOT
```
marikomedlock: Should this be empty? I think the SNAPSHOT suffix is only for the libs-snapshot-local Maven repository.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,80 @@
...
+#        if: always()
+#        uses: actions/upload-artifact@v1
+#        with:
+#          name: Test Reports
+#          path: build/reports/tests
+      - name: "Bump the tag to a new version"
+        uses: databiosphere/github-actions/actions/bumper@v0.0.3
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DEFAULT_BUMP: patch
```
marikomedlock, line 57: How do we bump minor or major version instead?

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,80 @@
...
+#        run: ./gradlew unitTest --scan
+#      - name: Run integration tests
+#        id: integration-test
+#        run: ./gradlew integrationTest --scan
+#      - name: Upload Test Reports
+#        if: always()
+#        uses: actions/upload-artifact@v1
+#        with:
+#          name: Test Reports
+#          path: build/reports/tests
```
marikomedlock: Nit: Remove or uncomment.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,80 @@
...
+      - name: "Bump the tag to a new version"
+        uses: databiosphere/github-actions/actions/bumper@v0.0.3
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DEFAULT_BUMP: patch
+          RELEASE_BRANCHES: master,develop
```
marikomedlock: How does this relate to the develop branch specified in the Checkout step above?

src/main/java/bio/terra/common/migirate/LiquibaseMigrator.java
```diff
@@ -0,0 +1,84 @@
...
+                changesetFile, lock.getId(), lock.getLockedBy(), lock.getLockGranted()));
+
+        // We can get into this state where one of the APIs is running migrations and gets shut down
+        // so that
+        // another API container can run. It will result in a lock that doesn't get released. This
+        // is similar
+        // to the problems we will have from deploying multiple containers at once that try to run
+        // migrations.
+        logger.warn("Forcing lock release");
+        liquibase.forceReleaseLocks();
```
marikomedlock, line 70: I don't completely follow this part. If two pods both try to run migrations, will the second one force release the lock that the first one took out? (I'm probably missing some context on how/when this gets called, so feel free to just point me somewhere I should look/read up. Thanks.)

.github/workflows/publish-release.yml
yonghaoy: I don't think there is a way to bump both major&minor versions. For major version update, I think we may have to manually update version in build.gradle.  The current workflow would be: 
1. PR to update versions in build.gradle
2. Create a release, and it will start from the major version.

Does that SG?  I feel that is okay for now to add one extra manually step to upgrade major version?

.github/workflows/publish-release.yml
yonghaoy: okay, removed it.  I leave it comment here it remind myself or other libraries: run test before publish.

.github/workflows/publish-release.yml
yonghaoy: That is for which branches will generate release tags:
https://github.com/DataBiosphere/github-actions/tree/master/actions/bumper#options

I don't think I will need master here anymore. Leave it here just in case other libraries copy-paste from here but forget to change this line...

src/main/java/bio/terra/common/migirate/LiquibaseMigrator.java
yonghaoy: This is the pattern in all existing services:
https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/upgrade/Migrate.java#L338
, https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/service/migrate/MigrateService.java
I don't know why we have this at beginning, but I assume this is related with the javadoc above? `"We expect MC Terra services to scale horizontally and to perform rolling upgrades of the running instances"`

.github/workflows/publish-release.yml
marikomedlock: I see, okay. Maybe we can just put this in the README then?

.github/workflows/publish-release.yml
marikomedlock: Okay, up to you. My slight preference is to just put a note in the README to change the branch to master if that's the "main" branch for the repository.

src/main/java/bio/terra/common/migirate/LiquibaseMigrator.java
marikomedlock: Okay, I'm not sure I follow that. But I understand you're just copying code to centralize it here, so this question should probably be directed to the groups that are using that code now.

marikomedlock (approved):
LGTM. One request is to update the README with more details about which part of the version number gets bumped when -- I think it's patch number on PR merge and minor/major number on build.gradle file version property change and GH release.

PR #4, opened 2020-12-16
[PF-276] add package for common exceptions and exception handler

yonghaoy:
Example of usage: https://github.com/DataBiosphere/terra-resource-buffer/pull/80


src/main/java/bio/terra/common/exception/AbstractGlobalExceptionHandler.java
```diff
@@ -0,0 +1,68 @@
...
+ *   ErrorReport generateErrorReport(Throwable ex, HttpStatus statusCode, List<String> causes) {
+ *     return new ErrorReport().message(ex.getMessage()).statusCode(statusCode.value()).causes(causes);
+ *   }
+ * }
+ * </pre>
+ */
+abstract class AbstractGlobalExceptionHandler<T> {
+  private final Logger logger = LoggerFactory.getLogger(AbstractGlobalExceptionHandler.class);
+
+  // -- Error Report - one of our exceptions --
```
wchamber: fix these comment styles to be javadoc ditto below

src/main/java/bio/terra/common/exception/AbstractGlobalExceptionHandler.java
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.MethodArgumentNotValidException;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.servlet.NoHandlerFoundException;
+
+import java.util.List;
+
+/**
+ * This module provides an abstract as a top-level exception handler for controllers. All exceptions that rise
+ * through the controllers are caught in this handler. It converts the exceptions into standard
```
wchamber: "`All exceptions that rise through the controllers are caught in this handler.`" That sounds like it happens magically, but I think that's aspirational. Let's change the phrasing so that it's clear what this base class does, vs what should be done by subclasses.

src/main/java/bio/terra/common/exception/BadRequestException.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.http.HttpStatus;
+
+import java.util.List;
+
+/**
+ * This base class has data that corresponds to the ErrorReport model generated from the OpenAPI
+ * yaml. The global exception handler auto-magically converts exceptions of this base class into the
+ * appropriate ErrorReport REST response.
+ */
+public class BadRequestException extends ErrorReportException {
```
wchamber: this javadoc comment doesn't seem to match this class. Ditto others. Let's have the same kind of comment for all the sub classes of ErrorReportException.

src/main/java/bio/terra/common/exception/InternalServerErrorException.java
```diff
@@ -0,0 +1,32 @@
+package bio.terra.common.exception;
+
+import org.springframework.http.HttpStatus;
+
+import java.util.List;
+
+/**
+ * This exception maps to HttpStatus.INTERNAL_SERVER_ERROR in the GlobalExceptionHandler.
+ */
+public abstract class InternalServerErrorException extends ErrorReportException {
```
wchamber: worth a comment on why this is an `abstract class` but none of the other subclasses of ErrorReportException are.

wchamber (changes requested):
mostly asking for improved comments. I know you're just copying what already exists, but now's the time to fix it up. thanks!

src/main/java/bio/terra/common/exception/AbstractGlobalExceptionHandler.java
yonghaoy: tried to revised the stuff. PTAL, thanks! The new javadoc is trying to get attention to {@link ExceptionHandler} some of answers are explained in that javadoc

src/main/java/bio/terra/common/exception/BadRequestException.java
yonghaoy: Done, not sure if that is accurate through.  PTAL

src/main/java/bio/terra/common/exception/InternalServerErrorException.java
yonghaoy: actually this shouldn't be abstract. I don't think we should block client using this directly...

yonghaoy (commented):
also added some test and improve github action
Thanks

.github/workflows/develop-push.yml
```diff
@@ -14,10 +14,10 @@ jobs:
         uses: actions/checkout@v2
         with:
           ref: develop
-      - name: Set up JDK 1.8
-        uses: actions/setup-java@v1
+      - name: Set up AdoptOpenJDK 11
+        uses: joschi/setup-jdk@v2
         with:
-          java-version: 1.8
+          java-version: 11
```
wchamber, line 20: Please try to keep unrelated changes in separate PRs. Update title and description as appropriate.

src/main/java/bio/terra/common/exception/AbstractGlobalExceptionHandler.java
```diff
@@ -0,0 +1,69 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.MethodArgumentNotValidException;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.servlet.NoHandlerFoundException;
+
+/**
+ * This class binds exception classes with Spring HTTP {@link ResponseEntity} by using {@link
+ * ExceptionHandler}. Subclasses implements {@link #generateErrorReport} to use their own
+ * ErrorReport defined in server's openapi yaml file, then HTTP responses can have the expected HTTP
+ * status code in defined in each exception classes.
```
wchamber: typo nit: `status code as defined in...`

src/test/java/bio/terra/common/exception/AbstractGlobalExceptionHandlerTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.List;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.springframework.http.HttpStatus;
+
+/** Test for {@link AbstractGlobalExceptionHandler} */
+@Tag("unit")
+public class AbstractGlobalExceptionHandlerTest {
+  TestExceptionHandler exceptionHandler = new TestExceptionHandler();
```
wchamber: nit: this member var and the static classes can & should be private.

wchamber (approved):

.github/workflows/develop-push.yml
yonghaoy: ah, sorry. The whole repo is just get setup. This is the first chunk code I added, and there are some problems I realized while I add this PR. So I just improve it in this PR. 
Updated PR comment.

PR #6, opened 2020-12-17
[no ticket] Fixe publishing error

yonghaoy:
In my last PR I added SpringBoot gralde plugin but that disable regular Jar and break publish package. 
Actually this is not a app, it is a library. So I disable bootJar and enable regular jar back


marikomedlock (approved):
LGTM

PR #7, opened 2020-12-22
[PF-318] Add a utility class to help execute database operation with retry support

yonghaoy:
I think Buffer Service needs this now.. So I just added it here.
Hi @gjuggler 
Not sure if package name/class name/ method name is accurate. 
Thanks!

### DIFF
--- a/src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
+++ b/src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
@@ -1,0 +1,47 @@
+package bio.terra.common.db;
+
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+
+/** Utilities for handling data base retry operations. */
+public class DatabaseRetryUtils {
+    private static final Logger logger = LoggerFactory.getLogger(DatabaseRetryUtils.class);
+
+    /** Executes a database operation and retries if retryable. */
+    public static <T> T executeAndRetry(
+            DatabaseExecute<T> execute, Duration retrySleep, int maxNumRetries)
+            throws InterruptedException {
+        int numRetries = 0;
+        while (numRetries < maxNumRetries) {
+            try {
+                return execute.execute();
+            } catch (DataAccessException e) {
+                System.out.println("~~~~~~~~~retry exception!!!!!!");
+                System.out.println(numRetries);
+                System.out.println(e.getLocalizedMessage());
+                if (!retryQuery(e)) {
+                    throw e;
+                }
+            }
+            ++numRetries;
+            TimeUnit.MILLISECONDS.sleep(retrySleep.toMillis());
+        }
+        throw new InterruptedException("Exceeds maximum number of retries.");
+    }
+
+    /** Returns {@code true} if that is retryable {@link DataAccessException}. */
+    public static boolean retryQuery(DataAccessException dataAccessException) {
+        return ExceptionUtils.hasCause(dataAccessException, RecoverableDataAccessException.class)
+                || ExceptionUtils.hasCause(dataAccessException, TransientDataAccessException.class);
+    }
+
+    /** How to execute this database operation. */
+    @FunctionalInterface
+    public interface DatabaseExecute<T> {
+        T execute();
+    }
+}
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
```diff
@@ -0,0 +1,47 @@
...
+        logger.info("Retrying time: {}", numRetries, e);
+      }
+      ++numRetries;
+      TimeUnit.MILLISECONDS.sleep(retrySleep.toMillis());
+    }
+    throw new InterruptedException("Exceeds maximum number of retries.");
+  }
+
+  /** Returns {@code true} if that is retryable {@link DataAccessException}. */
+  public static boolean retryQuery(DataAccessException dataAccessException) {
```
gjuggler: naming nit: shouldRetryQuery (more clearly aligns w/ a bool return type)

src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.dao.RecoverableDataAccessException;
+import org.springframework.dao.TransientDataAccessException;
+
+/** Utilities to execute database operations with retry support. */
+public class DatabaseRetryUtils {
+  private static final Logger logger = LoggerFactory.getLogger(DatabaseRetryUtils.class);
+
+  /** Executes a database operation and retries if retryable. */
+  public static <T> T executeAndRetry(
+      DatabaseExecute<T> execute, Duration retrySleep, int maxNumRetries)
```
gjuggler: naming nit: s/maxNumRetries/maxNumAttempts/?

If I'm reading correctly, with maxNumRetries=1, execution will only proceed once through before an InterruptedException is thrown, meaning zero retries were attempted. I think "attempts" as a parameter is less ambiguous. It's also more consistent with Spring Retry (https://www.baeldung.com/spring-retry)

src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
```diff
@@ -0,0 +1,47 @@
...
+      throws InterruptedException {
+    int numRetries = 0;
```
gjuggler: If you switch to "attempts" semantics for the parameter, it might make sense to switch to attempts here too for consistency, e.g. `int numAttempts = 1;`

src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
```diff
@@ -0,0 +1,47 @@
...
+    while (numRetries < maxNumRetries) {
+      try {
+        return execute.execute();
+      } catch (DataAccessException e) {
+        if (!retryQuery(e)) {
+          throw e;
+        }
+        logger.info("Retrying time: {}", numRetries, e);
```
gjuggler: Log message phrasing and semantics are a bit odd here. Maybe:

logger.info("Caught exception, retrying DB operation. Attempts so far: {}", numAttempts, e)

src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
```diff
@@ -0,0 +1,47 @@
...
+
+  /** Returns {@code true} if that is retryable {@link DataAccessException}. */
+  public static boolean retryQuery(DataAccessException dataAccessException) {
+    return ExceptionUtils.hasCause(dataAccessException, RecoverableDataAccessException.class)
+        || ExceptionUtils.hasCause(dataAccessException, TransientDataAccessException.class);
+  }
+
+  /** How to execute this database operation. */
+  @FunctionalInterface
+  public interface DatabaseExecute<T> {
```
gjuggler: Naming nit: maybe DatabaseOperation is more appropriate? The fact that your javadocs refer to this as the "database operation" suggests that may be the right class name.

src/test/java/bio/terra/common/db/DatabaseRetryUtilsTest.java
```diff
@@ -0,0 +1,65 @@
+package bio.terra.common.db;
```
gjuggler, line 1: Nice tests!

gjuggler (approved):
Thanks – this will be a really nice reusable piece of code, esp. as we align on serializable transactions as default across services.

One might argue that much of this could be achieved with some sort of Spring Retry use, but I think I prefer the simplicity & directness of this approach.

A few naming / semantic nits. Happy for this to go in once you've resolved to your satisfaction.

src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
yonghaoy: Done. I copied the code from [here](https://github.com/DataBiosphere/jade-data-repo/blob/ef3b53b57cd46e4a65309f5acf2030a31c8cc52f/src/main/java/bio/terra/common/DaoUtils.java#L87) and didn't think too much about the naming

PR #11, opened 2021-01-29
PF-281 Add a StairwayHook to facilitate tracing Stairway execution

wchamber:
Adds a StairwayHook to create OpenCensus tracing Spans for flight & step execution. With this hook, any service running stairway will get traces (as determined by sampling rate) for flight execution.

Example traces from my personal environment with WM integration https://github.com/DataBiosphere/terra-workspace-manager/pull/206:
- Trace for flight to create a workspace: https://console.cloud.google.com/traces/list?tid=7c5c400b07be45dcd770291bd579e065
- Trace for flight to create a google project: https://console.cloud.google.com/traces/list?tid=74608f25bc81f77d46c2e711ecd934f6, which is linked to the submission request: https://console.cloud.google.com/traces/list?tid=387c9d31f5a3ed68b6df3433554c020b 

Adds a local postgres instance for unit tests with Stairway. Run a postgres server in the github test action.


src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,205 @@
...
+
+  // Prefixes to use for Span names. Standard prefixes make it easier to search for all Spans of
+  // different types.
+  private static final String SUBMISSION_NAME_PREFIX = "stairway/submission/";
+  private static final String FLIGHT_NAME_PREFIX = "stairway/flight/";
+  private static final String STEP_NAME_PREFIX = "stairway/step/";
+
+  /** Map of Flight IDs to Flight Spans. */
+  // DO NOT SUBMIT: Consider whether we want to have a FlightHook to match StepHook instead of
+  // storing per Flight data in a global map. Note that the FlightHook would have to be able to make
```
ddietterich: FlightHook would be a straightforward extension.

src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,205 @@
...
+   */
+  public static Object serializedCurrentContext() {
+    return encodeContext(Tracing.getTracer().getCurrentSpan().getContext());
+  }
+
+  /** Store the current Span's {@link SpanContext} as the submission Span. */
+  public static void storeCurrentContextAsSubmission(FlightMap inputMap) {
+    inputMap.put(
+        SUBMISSION_SPAN_CONTEXT_MAP_KEY,
+        encodeContext(Tracing.getTracer().getCurrentSpan().getContext()));
```
ddietterich: Could use serializedCurrentContext

src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  /** A {@link StepHook} for creating Spans for each Step execution. */
+  private class TraceStepHook implements StepHook {
+    private Span stepSpan = null;
+    private Scope stepScope = null;
+
+    @Override
+    public HookAction startStep(FlightContext flightContext) {
+      Span flightSpan = flightSpans.get(flightContext.getFlightId());
```
ddietterich: Couldn't you get this in the constructor?

src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,205 @@
...
+      stepSpan.putAttribute("stairway/direction", AttributeValue.stringAttributeValue(flightContext.getDirection().toString()));
+      // Start the Scope of the Step Span's execution. We must remember to close the Scope at the
+      // end of the Step.
+      stepScope = tracer.withSpan(stepSpan);
+      return HookAction.CONTINUE;
+    }
+
+    @Override
+    public HookAction endStep(FlightContext flightContext) {
+      stepScope.close();
```
ddietterich: I'm not hip to scopes. I notice you don't use one for flights. How come?

ddietterich (commented):
A few trivial comments. I don't know the opencensus objects well, so can't comment on correctness there.

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: Any preference for or against adding it?

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: done

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: done

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: The scope changes what the "current span" is, e.g. what `Tracing.getTracer().getCurrentSpan()` returns (implemented with ThreadLocals I believe). It's important for the step execution to run within a scope so that each step can add to the current span and get the right results. As Stairway doesn't do any span additions by itself, changing what the current span is for the flight isn't necessary, but also isn't harmful (as long as `startFlight`/`endFlight` are called on the same thread as `startStep`/`endStep`. As is, I'd have to add a pair or second map of `flightScopes` or add my own value class to store the Flight Scope in addition to the Flight Span, so it didn't seem worth it. Less work if we had a `FlightHook`.
Tried to add some more explanation to the comments.

src/main/java/bio/terra/common/tracing/TracingHook.java
ddietterich: If it would be useful, let's add it right away before we get more hook code written!
It would be small work. And upward compatible

yonghaoy:
There was a open question in common library doc about which package this should live: 
`bio.terra.common.stairway` vs `bio.terra.common.tracing`. 
@ddietterich also left a comment here: 
https://docs.google.com/document/d/1Q6F7bOeG_FoEpskYxQ7NqgknVmca-uMVFC_wz4MLw4U/edit?disco=AAAALOy1v7c

yonghaoy:
I only see the GreateGoogleProjectStep, but where are the following step? Like this one `SetProjectBillingStep`
https://console.cloud.google.com/traces/list?tid=c79ef9d9b94cd8d64d7ff499efdf4986&project=terra-kernel-k8s&pageState=(%22traceIntervalPicker%22:(%22groupValue%22:%22PT12H%22,%22customValue%22:null))

src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,214 @@
...
+    return Base64.getEncoder()
+        .encodeToString(
+            Tracing.getPropagationComponent().getBinaryFormat().toByteArray(spanContext));
+  }
+
+  /**
+   * Decode the SpanContext from a base64 string using the binary format. Works with {@link
+   * #encodeContext(SpanContext)}.
+   */
+  private static SpanContext decodeContext(String encodedContext) {
```
yonghaoy: Can we use `SpanContext.class` in flight input map?
```
SpanContext spanContext =
        flightContext.getInputParameters().get(SPAN_CONTEXT, SpanContext.class); 
```

src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,214 @@
...
+    // Check the working map for a submission span to link for the Flight.
+    encodedContext =
+        flightContext.getWorkingMap().get(SUBMISSION_SPAN_CONTEXT_MAP_KEY, String.class);
+    if (encodedContext != null) {
+      return decodeContext(encodedContext);
+    }
+    // Create a new span to use as a submission span to link for the Flight Spans.
+    Span submissionSpan =
+        tracer
+            .spanBuilderWithExplicitParent(
```
yonghaoy: We already serelize/deserelize spanContext in flight Map.
if we use [spanBuilderWithRemoteParent](https://www.javadoc.io/doc/io.opencensus/opencensus-api/0.17.0/io/opencensus/trace/Tracer.html#spanBuilderWithRemoteParent-java.lang.String-io.opencensus.trace.SpanContext-)
Can we also get correct span even in flight recover case?

yonghaoy (commented):
Thanks

src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,214 @@
...
+  @Override
+  public HookAction startFlight(FlightContext context) {
+    SpanContext submissionContext = getOrSubmissionContext(context);
+    // Start the Flight Span. We must remember to close the Flight Span at the end of the Flight's
+    // current run.
+    // We do not currently change the scope to use the Flight Span as the current span. We mostly
+    // only care that the code executed within a step is within the scope of one of our spans.
+    Span flightSpan =
+        Tracing.getTracer()
+            .spanBuilder(FLIGHT_NAME_PREFIX + context.getFlightClassName())
```
gjuggler: The fully-qualified class name is a bit verbose for the top-level span names. It might make sense to use just the class name itself here, e.g.

stairway/flight/bio.terra.workspace.service.workspace.flight.CreateGoogleContextFlight

becomes

stairway/flight/CreateGoogleContextFlight

I don't think this is too much of a loss, since traces are generally searched for & viewed in the context of a service where the uniqueness of the package path will be more or less implied. And we still have the full classname as a label.

Either a simple regexp or something like https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/ClassUtils.html#getShortCanonicalName-java.lang.String- should work.

src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   * Creates a Flight Span for the current run of the Flight. Note that the same Flight may be run
+   * multiple times by Stairway.
+   */
+  @Override
+  public HookAction startFlight(FlightContext context) {
+    SpanContext submissionContext = getOrSubmissionContext(context);
+    // Start the Flight Span. We must remember to close the Flight Span at the end of the Flight's
+    // current run.
+    // We do not currently change the scope to use the Flight Span as the current span. We mostly
```
gjuggler: My understanding is that we're using what OpenCensus refers to as "manual context propagation" here ([ref](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/Tracer.java#L52)), by actively attaching each step span to this parent span. May be better to call that out, e.g. "Rather than use OpenCensus' scope-based span linkage, we manually attach step spans to this parent span via the stepFactory method."

src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,214 @@
...
+  private final Map<String, Span> flightSpans = new ConcurrentHashMap<>();
+
+  /** The {@link FlightMap} key for the submission Span's context. */
+  public static final String SUBMISSION_SPAN_CONTEXT_MAP_KEY = "opencensusTracingSpanContext";
+
+  /**
+   * Serialize the current Span's {@link SpanContext}. To be used to store a submission Span for the
+   * Flight with {@link #SUBMISSION_SPAN_CONTEXT_MAP_KEY}.
+   */
+  public static Object serializedCurrentContext() {
```
gjuggler: Does this need to be public? It may be better to only make the storeCurrentContext method accessible, and/or just wrap this method into that one if it's only used there.

(And, naming nit: prefer verb-style names, e.g. serializeCurrentTracingContext)

src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,214 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * A {@link StairwayHook} to add support for tracing execution of Stairway flights with OpenCensus.
+ *
+ * <p>This hook uses Spans at 3 different levels. From lowest to highest:
+ * <li>A Span around the execution of each Step, one Span per Step execution. All of the {@link
+ *     Step#doStep(FlightContext)} {@link Step#undoStep(FlightContext)} will be executed in the
+ *     context of this Span.
+ * <li>A Span around the execution of each Flight, one Span per {@link Flight#run()}. Note that a
```
gjuggler: This part was a little conceptually jarring to me.

The most ideal output, from a naive developer's perspective, would be for there to be one top-level span for the entire flight progression, regardless of pause and resume events. Identifying pause / resume events with https://opencensus.io/tracing/span/time_events/annotation/ would be icing on the cake.

Without going too overboard, I wonder if there's some way to handle this. Maybe you could store the SpanContext for the top-level Flight span within the working map, and do a get-or-create in the initial startFlight call?

To some extent, this would feel like what's already being done in the case where a manual submission context span needs to be created. Maybe there's ultimately not a need for both? i.e. right now we conditionally create a Submission span to act as the parent of N flight-execution spans, but maybe we can always create one Flight span (optionally linked to a service-provided submission span) to act as the parent of N slight-step spans.

src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,214 @@
+package bio.terra.common.tracing;
+
+import bio.terra.stairway.*;
```
gjuggler: Not sure if we've been strict about this at all, but Google's style guide disallows wildcard imports ([ref](https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports)). CH was careful about this in the AoU repo... I'm ambivalent, especially since IDEs seem to reach for wildcards sometimes.

src/main/java/bio/terra/common/tracing/TracingHook.java
```diff
@@ -0,0 +1,214 @@
...
+    @Override
+    public HookAction startStep(FlightContext flightContext) {
+      // Start the Step Span. We must remember to close the Span at the end of the Step.
+      stepSpan =
+          Tracing.getTracer()
+              .spanBuilderWithExplicitParent(
+                  STEP_NAME_PREFIX + flightContext.getStepClassName(), flightSpan)
+              .startSpan();
+      stepSpan.putAttribute(
+          "stairway/flightId", AttributeValue.stringAttributeValue(flightContext.getFlightId()));
```
gjuggler: nit: consider adding stepIndex too?

A serialized version of the input map and working map would be cool too for debugging, but I suppose those may be large and/or contain sensitive values so we shouldn't be persisting them outside of the application db.

gjuggler (commented):
Really nice setup! One high-level comment / question around submission vs flight spans, otherwise just nits.

src/main/java/bio/terra/common/tracing/TracingHook.java
gjuggler: Ah, I see that this is indeed used in https://github.com/DataBiosphere/terra-workspace-manager/pull/206/files – makes sense.

src/main/java/bio/terra/common/tracing/TracingHook.java
yonghaoy: I just setup it correctly when I onboard AoU. We missed Google-Java-Format in our intellij. 
Instruction: https://docs.google.com/document/d/1DtESBapEzvuti7xODTFPHorwmLM7LybF-6D5lhbIkLU/edit#heading=h.vj6yd

wchamber:
> I only see the GreateGoogleProjectStep, but where are the following step? Like this one `SetProjectBillingStep`
> https://console.cloud.google.com/traces/list?tid=c79ef9d9b94cd8d64d7ff499efdf4986&project=terra-kernel-k8s&pageState=(%22traceIntervalPicker%22:(%22groupValue%22:%22PT12H%22,%22customValue%22:null))

That's because that flight failed to execute. It got stuck creating the project due to the ongoing permissions snafu in the WM move to RBS. So it never got past trying to create the google project to any of the later steps.

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: Nope! To have a single span for the entire flight progression, we would have to be able to serialize and resume that span in case it hops between Stairway instances. This is explicitly not possible through the public APIs of the OpenCensus library i.e. you cannot create a Span from its own SpanContext. You can only create a new Span. See the methods on https://javadoc.io/static/io.opencensus/opencensus-api/0.28.3/io/opencensus/trace/Tracer.html and note that their are no public constructors of SpanBuilder or Span.
A Span is local to a process. We can propagate SpanContexts to create child Spans (when the child *will not* outlive the parent Span) or create Spans with a link to the parent Span (when the child *may* outlive the parent Span.)

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: There's good reasons to not use wildcard imports, but I'm not really interested in fighting my IDE much less everyone else's IDE in code review. IF we have a hard desire to stop wildcard imports and can get everyone to change their IDE settings/have a pre-submit to disallow them so they don't get submitted, I will happily jump on board. But without broad consensus/enforcement, I'm not interested in trying.

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: Yep. IMO it's a bit annoying how this iteration of JobService obscures the FlightMap. I'd like to have full control in this class what gets stored in the FlightMap and how, but instead I have to make each key/value public.
Updating the fn name.

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: Sorry, I'm not sure what you mean.
We do not want the Flight Spans to be the children of the Submission span because the Flight Spans may outlive the Submission Span. Child spans must be completed before their parent spans. With the parent *link*, we can have the "child linked" spans outlive their "parent linked" spans. (Terminology around linkage is confusing, but that's what they have. https://opencensus.io/tracing/span/link/)

I don't know what you mean by the flight recover case. The same Flight Span is in use when the Flight is undone on error.

src/main/java/bio/terra/common/tracing/TracingHook.java
ddietterich: It is a trivial IntelliJ configuration to stop getting wildcard imports. That was the norm for TDR and I think it should be the norm here as well. With that config in place, we did not need any pre-submit rule.

src/main/java/bio/terra/common/tracing/TracingHook.java
gjuggler: Thanks for the explanation – I think I understand now!

Thinking from a user's perspective on the Trace visualization page, it would still be really nice to know whether a given Flight span represents the complete vs. partial execution of a given flight. To whatever extent we can surface that information from within a hook (or bubble up some feature requests to Stairway to propagate that context), that would be nice... but obviously not mission critical, especially if this context is at least captured in the correlated logs.

And thinking about potential future analytical use cases (e.g. digging into span data via bigquery), getting the overall runtime of a flight won't be as simple as looking at the duration of a span w/ that classname. You'd have to do something like min/max timestamp of all spans with the same classname *and* the same flight ID. Or, otherwise perhaps exclude spans that don't encapsulate an entire flight, since we expect that to be the minority... though this is only possible if we have some annotations that let us filter out these spans.

wchamber: Added the flight status at the end of the flight span.
Looking for [traces with the matching flight Id](https://console.cloud.google.com/traces/list?tid=c79ef9d9b94cd8d64d7ff499efdf4986&project=terra-kernel-k8s&pageState=(%22traceIntervalPicker%22:(%22groupValue%22:%22P4D%22,%22customValue%22:null),%22traceFilter%22:(%22chips%22:%22%255B%257B_22k_22_3A_22LABEL_3AflightId_22_2C_22t_22_3A10_2C_22v_22_3A_22_5C_22971c53ef-9f47-449a-9511-2960577748df_5C_22_22%257D%255D%22))) works pretty well for finding everything having to do with the flight (for spans that are not still in progress.)

I think if we're trying to do metrics analytics, we should have a metrics plugin. (I think we should have a metrics plugin regardless for recording flight & step durations, retry rates, error rates, but YMMV.)
I think of traces as being more for interactive debugging. Adding hooks for the "complete execution" of the flight would also be helpful.

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: sgtm. I think getShortClassName is right here since Stairway gets the class name and not its canonical name.
https://github.com/DataBiosphere/stairway/blob/28eaca8fff4967a7149442b4111c00e9d42ebbcc/src/main/java/bio/terra/stairway/Flight.java#L53

left the full names for the attributes.

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: done

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: thanks, done.
(With a `FlightHook` I think I'll switch it to setting the context on the flight span and relying on the automatic propagation of the context for the step span)

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: That relies on the jackson object mapper being able to serialize/deserialize correctly. And that being backwards compatible for future versions of OpenCensus. I'd rather use the OpenCensus provided & supported serialization, even if it's opaque in the serialized FlightMap.

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: per offline, Dan to tackle adding this to Stairway. Will update here once ready.

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: done.

wchamber (commented):
Replied to comments.
Still want to add unit tests with a local Stairway instance as well as the FlightHook changes.

src/main/java/bio/terra/common/tracing/TracingHook.java
yonghaoy: Sorry for the confusion, the recovery I mean is this scenario: Flight running, app is redeployed. Then we lost the original `flightSpan`. Do you think it make sense if we reconstruct the `remoteParentSpanContext` ourself.
Just my through, totally optional.

yonghaoy (approved):
Thanks Will!

gjuggler (approved):
Nice tests!

src/main/java/bio/terra/common/tracing/TracingHook.java
wchamber: We can't resume a Span, even with a SpanContext. So if the app crashes ungracefully, and we have dangling Spans, they will not get exported. See my longer answer to Greg about why we can't have a single Span for the entire Flight's lifetime across Stairway instances.

PR #13, opened 2021-02-03
PF-281 Fix the publish workflow to start up Postgres for its unit test runs.

wchamber:
Broken in PR #11 
Maybe later we should investigate having the publish workflow use the test workflow instead of copying its test executions.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -8,11 +8,27 @@ env:
 jobs:
   publish_release:
     runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:12.3
+        env:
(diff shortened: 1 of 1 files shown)


gjuggler (approved):

gjuggler:
> Broken in PR #11
> Maybe later we should investigate having the publish workflow use the test workflow instead of copying its test executions.

Is this about gaining some level of code sharing between the workflows? I looked at this briefly when Isaac introduced the "matrix variables approach" in WSM, but GitHub's support for any level of code sharing seemed unfortunately quite lacking.

wchamber:
Some kind of code sharing or workflow re-use. Or say, only run this workflow after the test workflow has passed on develop. I didn't look at what matrix variables are.

PR #15, opened 2021-02-03
[PF-278] First pass at Kubernetes common library

tnadler:


build.gradle
```diff
@@ -23,7 +23,7 @@ repositories {
 
 group 'bio.terra'
 version '0.0.15-SNAPSHOT'
-sourceCompatibility = JavaVersion.VERSION_11
+sourceCompatibility = JavaVersion.VERSION_1_8
```
jaycarlton, line 26: Maybe this change by itself should be its own PR, round of testing, and version change.

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
```diff
@@ -0,0 +1,184 @@
...
+  private static final int WATCH_INITIAL_WAIT = 5;
+  private static final int WATCH_MAX_WAIT = 30;
+
+  private final Logger logger = LoggerFactory.getLogger(KubePodListener.class);
+  private final KubeService kubeService;
+  private final String namespace;
+  private final String podNameFilter;
+  private final Stairway stairway;
+  private Exception exception;
+  private final Map<String, Boolean> podMap; // pod name; true means running; false means deleted
```
jaycarlton: Please use a `PodState` enum instead of a boolean for readability and expandability.

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * @param podName Name of the Kubernetes pod we are running in. If we are not in a pod, this
+   *     defaults to a constant string in application properties. We cannot easily find our own pod
+   *     name, so components must plumb this in using settings in the helm charts.
+   * @param inKubernetes Used to denote that we are running in the Kubernetes environment. The
+   *     service can be called and provides reasonable answers when inKubernetes is false.
+   * @param podNameFilter Filter to apply to pod names returned by Kubernetes API to know which pods
+   *     we care about.
+   */
+  public KubeService(String podName, boolean inKubernetes, String podNameFilter) {
```
yonghaoy, line 64: I am not 100% sure thow to use podNameFilter and podName. 
Draft a PR: https://github.com/broadinstitute/terra-helm/pull/286/files

```
podName: metadata.name
podNameFilter: the one in line 25?
```
Can you help to take look to see if my understanding is correctly. 
Thanks

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    try {
+      CoreV1Api api = makeCoreApi();
+      V1PodList list =
+          api.listNamespacedPod(namespace, null, null, null, null, null, null, null, null, null);
+      for (V1Pod item : list.getItems()) {
+        if (item.getMetadata() != null) {
+          String podName = item.getMetadata().getName();
+          if (StringUtils.contains(podName, podNameFilter)) {
```
yonghaoy, line 103: In the original TDR code, you have `getPods()` `getApiPods()`, the 2nd one uses `podNameFilter`. 
Do we need to keep the original `getPods()`, that does not have the filter?

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -0,0 +1,195 @@
...
+    return true;
+  }
+
+  public int getActivePodCount() {
+    if (podListener != null) {
+      return podListener.getActivePodCount();
+    }
+    int defaultPodCount = 1;
+    logger.info("KubeService ActivePodCount - default val: {}", defaultPodCount);
+    return defaultPodCount;
```
yonghaoy: Does this always return 1?

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+  }
+
+  private String readFileIntoString(String path) {
+    try {
+      byte[] encoded = Files.readAllBytes(Paths.get(path));
+      return new String(encoded, StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      logger.error("Failed to read file: " + path + "; ", e);
+      return null;
```
yonghaoy: Client may not be aware `getNamespace()` may return null. Feel it is possible to cause NPE in client code. Suggest to use `Optional`, or throw exception there?

yonghaoy (commented):
Thanks! Few questions!

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
```diff
@@ -0,0 +1,184 @@
...
+    logger.info("KubePodListener starting");
+
+    exception = null;
+    int consecutiveRetryCount = 0;
+    int retryWait = WATCH_INITIAL_WAIT;
+    while (true) {
+      try {
+        ApiClient client = Config.defaultClient();
+        // infinite timeout
+        OkHttpClient httpClient =
```
jaycarlton, line 93: The names here are confusing me. Maybe inline `httpClient`.

src/main/java/bio/terra/common/kubernetes/KubeService.java
ddietterich: Only if we are not running under Kubernetes

ddietterich: No. It returns 1 if `inKubernetes` is false.

src/main/java/bio/terra/common/kubernetes/KubeService.java
ddietterich: Your setup of `podName` and `inKubernetes` look right to me. That is what TDR does [here](https://github.com/broadinstitute/datarepo-helm/blob/master/charts/datarepo-api/templates/api-deployment.yaml)

In TDR, the podNameFilter is a constant `datarepo-api` because that substring is used  for the main DRmanager code in all deployments. So long as all Buffer Service pods in a deployment have `{{ .Values.name }}-service` as part of their name (and no other pod has it as part of its name), then it should work.

src/main/java/bio/terra/common/kubernetes/KubeService.java
ddietterich: I am not finding the `getPods()` you are talking about, so not understanding the question. I do not have a use case for exposing the whole pod list.

Now that I look at it, I have not done a good job making things package-private. I think I should do one more interface revision.

src/main/java/bio/terra/common/kubernetes/KubeService.java
yonghaoy: Did you push you commit to address this?

ddietterich: Not yet. I should have this updated by days end.

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
```diff
@@ -0,0 +1,184 @@
...
+            client.getHttpClient().newBuilder().readTimeout(0, TimeUnit.SECONDS).build();
+        client.setHttpClient(httpClient);
+        Configuration.setDefaultApiClient(client);
+
+        CoreV1Api api = new CoreV1Api();
+        try (Watch<V1Namespace> watch =
+            Watch.createWatch(
+                client,
+                api.listNamespacedPodCall(
```
jaycarlton, line 102: Is there a way to make an argument object that has a builder? This is a lot of nullable arguments.

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
```diff
@@ -0,0 +1,184 @@
...
+                try {
+                  logger.info("Attempting clean up of deleted stairway instance: " + podName);
+                  stairway.recoverStairway(podName);
+                  Boolean deletedPodValue = podMap.get(podName);
+                  if (deletedPodValue != null && deletedPodValue) {
+                    logger.info("Deleted api pod: " + podName);
+                    podMap.put(podName, false);
+                  }
+                } catch (DatabaseOperationException | StairwayExecutionException ex) {
+                  logger.error("Stairway recoveryStairway failed to recovery pod: " + podName, ex);
```
jaycarlton: *recover

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
```diff
@@ -0,0 +1,184 @@
...
+      }
+      logger.info("KubePodListener consecutive retry: " + consecutiveRetryCount);
+    }
+  }
+
+  Exception getException() {
+    return exception;
+  }
+
+  Map<String, Boolean> getPodMap() {
```
jaycarlton: If you don't want to make this enum-valued, then just a set of running pods would work I would think.

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -0,0 +1,190 @@
...
+@SuppressFBWarnings(
+    value = "DMI_HARDCODED_ABSOLUTE_FILENAME",
+    justification = "The K8s namespace file is a valid absolute filename")
+public class KubeService {
+  private static final Logger logger = LoggerFactory.getLogger(KubeService.class);
+
+  // Location in the container where Kubernetes stores service account and
+  // namespace information. Kubernetes mints a service account for the container (pod?)
+  // that can be used to make requests of Kubernetes.
+  static final String KUBE_DIR = "/var/run/secrets/kubernetes.io/serviceaccount";
```
jaycarlton, line 43: should this be `KUBE_SA_DIR`?

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -0,0 +1,190 @@
...
+            + podName);
+  }
+
+  /**
+   * Get a list of the API pods from Kubernetes. It is returned as a set to make probing for
+   * specific names easy.
+   *
+   * @return set of pod names containing the podNameFilter string; null if not in kubernetes
+   */
+  public Set<String> getPodList() {
```
jaycarlton: name this `getPodSet()` or `getPods()`

yonghaoy (approved):

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -0,0 +1,190 @@
...
+    Set<String> pods = new HashSet<>();
+    if (!inKubernetes) {
+      return pods;
+    }
+
+    try {
+      CoreV1Api api = makeCoreApi();
+      V1PodList list =
+          api.listNamespacedPod(namespace, null, null, null, null, null, null, null, null, null);
```
jaycarlton, line 99: Do we control this api? Can we overload `listNamespacePod`?

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -0,0 +1,190 @@
...
+  public void startPodListener(Stairway stairway) {
+    if (inKubernetes) {
+      podListener = new KubePodListener(this, stairway, namespace, podNameFilter);
+      podListenerThread = new Thread(podListener);
+      podListenerThread.start();
+    }
+  }
+
+  public int getActivePodCount() {
+    if (podListener != null) {
```
jaycarlton, line 128: I'd use `Optional.ofNullable().map().orElse()` here.

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -0,0 +1,190 @@
...
+
+  /**
+   * Stop pod listener flips the isShutdown flag so we don't do it more than once. It tries to join
+   * the pod listener thread.
+   *
+   * @param timeUnit time unit to wait for the listener thread to stop
+   * @param joinWait time in the time unit to wait
+   * @return true if we joined the listener thread; false if that timed out
+   */
+  public boolean stopPodListener(TimeUnit timeUnit, long joinWait) {
```
jaycarlton, line 144: `Duration` is easier to work with.

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  // Method for pod listener to test shutdown state
+  boolean isShutdown() {
+    return isShutdown.get();
+  }
+
+  private CoreV1Api makeCoreApi() {
+    try {
+      ApiClient client = ClientBuilder.cluster().build();
```
jaycarlton, line 174: This is confusing me. You make one kind of client and then return a different one?

jaycarlton (approved):
is there a way to write unit tests for this code?

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
ddietterich: Yes, the Kubernetes Java client API is crazy with nulls.

src/main/java/bio/terra/common/kubernetes/KubeService.java
ddietterich: No, we do not control the api.

src/main/java/bio/terra/common/kubernetes/KubeService.java
ddietterich: I would never use that. Why is everyone so twitchy about comparing to null? What is the benefit of running through 4 methods to do the same thing?

src/main/java/bio/terra/common/kubernetes/KubeService.java
ddietterich: This expression is a common idiom.

src/main/java/bio/terra/common/kubernetes/KubeService.java
ddietterich: This is how the K8s API manages its different features. You build different sub-clients for each thing you want to do. So there are alpha clients and beta clients. All kinds of clients.

ddietterich:
@jaycarlton I do not know a useful way to unit test this. Its entire purpose is interacting with Kubernetes.
I have been testing it by building data repo with it and running the repo integration tests that run under Kubernetes.

src/main/java/bio/terra/common/kubernetes/KubeService.java
jaycarlton: It's more deliberate, and a higher level of abstraction. There's no need to branch, and you're very clear about what you're getting and that it might not be there and what the default is.

`if` is basically deprecated as of Java 8.

PR #16, opened 2021-02-03
PF-281 Try to fix the javadoc so that we can publish to artifactory.

wchamber:
https://github.com/DataBiosphere/terra-common-lib/runs/1824254120?check_suite_focus=true
I don't really know of a good way to verify this works before merging. My `./gradlew javadoc` works fine either way.

### DIFF
--- a/src/main/java/bio/terra/common/stairway/TracingHook.java
+++ b/src/main/java/bio/terra/common/stairway/TracingHook.java
@@ -23,22 +23,25 @@ import org.apache.commons.lang3.ClassUtils;
  * A {@link StairwayHook} to add support for tracing execution of Stairway flights with OpenCensus.
  *
  * <p>This hook uses Spans at 3 different levels. From lowest to highest:
- * <li>A Span around the execution of each Step, one Span per Step execution. All of the {@link
- *     Step#doStep(FlightContext)} {@link Step#undoStep(FlightContext)} will be executed in the
- *     context of this Span.
- * <li>A Span around the execution of each Flight, one Span per {@link Flight#run()}. Note that a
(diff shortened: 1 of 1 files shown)


gjuggler:
> https://github.com/DataBiosphere/terra-common-lib/runs/1824254120?check_suite_focus=true
> I don't really know of a good way to verify this works before merging. My `./gradlew javadoc` works fine either way.

Ah, probably some battle between JDK versions or flavors. I wonder if we can force our actions to have a certain version, or print out more about the execution environment. (I've probably mentioned this before, but AoU's solution to this has been to wrap every dev command into a Dockerized tool execution. This guarantees consistency between devs and between local & CI executions, but also has a ton of overhead and drawbacks. I'd only want to reach for that if we start seeing these things become a serious drag on velocity, which... I don't think has happened yet? I don't have a great sense of how heavy this friction is.)

Semi-serious question: what purpose does Javadoc serve these days? Is it actually consumed in some meaningful way when an artifact is pulled into another repository? Is it only useful if the source isn't also included?

I've had some annoyance where Santa has killed my Gradle build because I haven't upvoted my local version of the javadoc binary. I wouldn't complain if we were to disable it for most of our Gradle builds.

gjuggler (approved):

wchamber:
re: javadoc's useful ness, I browse javadoc on various public packages. A consistent format is also helpful for tooling. I find it easier to navigate at a high level than scrolling through the source code. So the unsatisfying answer, is documentation purposes. Not that I know where ours ends up today. 

I don't have Santa problems because I have a gLinux laptop, so I can't speak to that. This is the first time it's caused me problems. And they're rather smaller than our self inflicted publish workflow issues.

PR #17, opened 2021-02-03
[PF-432] Fix javadoc

ddietterich:
Different javadoc versions have different complaints. I _think_ I have this fixed.

I also copied the .gitignore from WM to this library to hide build directories and such.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,38 @@
+HELP.md
+
+### Gradle ###
+.gradle
+**/build/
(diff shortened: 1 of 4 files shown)


yonghaoy (approved):

tnadler:
Drive by comment - when I was pushing to this directory I noticed my git access token needed to have workflow permission, which I had not seen on other repos. Is that WAI?

PR #19, opened 2021-02-11
[PF-405] Add a common logging package with full GCP support & per-request logs

gjuggler:
This PR implements each of the core logging features described in the PF-405 ticket:

- A library-provided logback.xml for consistent log config across service repos.
- A custom JsonLayout class supporting log-trace correlation, HttpRequest context info, and arbitrary structured logging.
- Servlet hooks to generate a log per API request (mimicking the behavior you get "for free" with App Engine Standard)
- A requestId filter that detects 'X-Request-Id' headers and/or generates a new requestId and inserts into MDC.
- Support for a 'human-readable-logging' Spring profile for local server debugging.

Plus some bonus integration tests to prove it works 😃 

I tried to pay close attention to minimizing the amount of config and boilerplate needed within a service repo. At this point, only two touch points are required, both in the 'Main.java' class of a service:
1. Add `bio.terra.common.logging` to the `@ComponenScan` annotation
2. Add the `LoggingInitializer` class to the Spring application initializers list. (This could technically be done by auto-config, but logging is initialized very early in the application lifecycle. Adjusting the log layout from an initializer ensures that all logs follow a consistent format, rather than switching midway through app initialization.)

For the most part, this PR is a consolidation and simplification of what WSM already has done. However, the new `GoogleJsonLayout` class fixes some bugs in the existing `org.springframework.cloud.gcp.logging.StackdriverJsonLayout.java` class, including a more reliable pattern for injecting `trace` and `span` IDs.

Having trace-linked logs becomes immediately apparent in Google's tracing detail view, where we can now see all log messages laid out in their tracing context:

Screen Shot 2021-02-08 at 11 19 28 PM

*Open questions*
* This PR opens an interesting, but slightly fraught, pattern of applications producing structured logs by sending JsonObject payloads as part of the call to `log.info(...)`. It's a really interesting option for minimizing friction to logging structured data... but we don't have concrete use cases in mind, and it could potentially be easily misused. My instinct is to not block this off entirely, but not lean too heavily into it yet.
* This PR leaves basically no logging options open to the service. If we end up seeing that services, or operational environments, have differing needs re: logging formats or verbosity, we may need to work harder to make improvements to the core library. This is probably a good thing, though – consistency by default, with some activation energy for special-casing services.


build.gradle
```diff
@@ -22,26 +22,42 @@ group 'bio.terra'
 version '0.0.11-SNAPSHOT'
 sourceCompatibility = JavaVersion.VERSION_11
```
ddietterich: I just set the library back to 8 so that TDR can use it.
Maybe pull the latest and make sure everything still works

build.gradle
```diff
@@ -125,7 +141,7 @@ task unitTest(type: Test) {
 }
 
 java {
-    withJavadocJar()
+//     withJavadocJar()
```
ddietterich: Why remove?

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -0,0 +1,227 @@
...
+ * Google Cloud Logging system.
+ *
+ * <p>See the LoggingTest integration test for more details on the output structure, including
+ * concrete examples of field values.
+ *
+ * <p>The output JSON relies heavily on Google Cloud's interpretation of structured JSON logs,
+ * including so-called "special fields in JSON payloads". See
+ * https://cloud.google.com/logging/docs/structured-logging#special-payload-fields for more details.
+ *
+ * <p>A goal of this class is to product maximally useful logs for operating Terra services on
```
ddietterich: nit: product -> produce

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -0,0 +1,227 @@
...
+  private ThrowableProxyConverter throwableProxyConverter;
+
+  public GoogleJsonLayout(ConfigurableApplicationContext applicationContext) {
+    this.applicationContext = applicationContext;
+    this.throwableProxyConverter = new ThrowableProxyConverter();
+    this.throwableProxyConverter.setOptionList(Collections.singletonList("full"));
+
+    // Configure the superclass.
+    this.appendLineSeparator = true;
+    setJsonFormatter(new Gson()::toJson);
```
ddietterich: Probably doesn't matter, but we are using Jackson most everywhere. Is it worth using it here instead of Gson?

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  @Override
+  public void start() {
+    super.start();
+    throwableProxyConverter.start();
+
+    // this.projectId = ServiceOptions.getDefaultProjectId();
```
ddietterich: Should this be in or out?

src/main/java/bio/terra/common/logging/LoggingConfig.java
```diff
@@ -0,0 +1,38 @@
...
+
+/** */
+@Configuration
+public class LoggingConfig implements WebMvcConfigurer {
+
+  Logger log = LoggerFactory.getLogger(LoggingConfig.class);
+
+  // Propagate or generate a new requestId at the beginning of our filter chain.
+  @Bean
+  @Order(0)
```
ddietterich, line 43: Maybe comment why the `@Order` is needed. It isn't obvious to me from looking at the code.

src/main/java/bio/terra/common/logging/LoggingInitializer.java
```diff
@@ -0,0 +1,52 @@
...
+    if (Arrays.stream(environment.getActiveProfiles()).anyMatch("human-readable-logging"::equals)) {
+      System.out.println("Human-readable logging enabled, skipping Google JSON layout");
+      // propertySources.addLast(
+      //     new MapPropertySource(
+      //         "logging-properties", Map.of("terra.common.logging.human-readable-format",
+      // "true")));
+      return;
+    }
+
+    // propertySources.addLast(
```
ddietterich: More commented out code

src/main/java/bio/terra/common/logging/LoggingInitializer.java
```diff
@@ -0,0 +1,52 @@
...
+  public static final String TERRA_APPENDER_NAME = "terra-common";
+
+  @Override
+  public void initialize(ConfigurableApplicationContext applicationContext) {
+    ch.qos.logback.classic.Logger logbackLogger =
+        (ch.qos.logback.classic.Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
+    ConfigurableEnvironment environment = applicationContext.getEnvironment();
+    // MutablePropertySources propertySources = environment.getPropertySources();
+
+    if (Arrays.stream(environment.getActiveProfiles()).anyMatch("human-readable-logging"::equals)) {
```
ddietterich: Is there a way to use a config property instead of a profile?

Ah, maybe that is what all of the commented code is. So maybe the answer is no?

src/main/java/bio/terra/common/logging/RequestIdFilter.java
```diff
@@ -0,0 +1,53 @@
...
+ * generated randomly.
+ *
+ * <p>The requestId is added to the MDC context and set as a response header.
+ */
+public class RequestIdFilter implements Filter {
+
+  public static final String REQUEST_ID_HEADER = "X-Request-ID";
+  public static final String REQUEST_ID_MDC_KEY = "requestId";
+
+  private final Hashids hashids = new Hashids("requestIdSalt", 8);
```
ddietterich, line 30: Why Hashid here? Why do we need a special library to turn a random int into a string?

build.gradle
```diff
@@ -25,29 +25,43 @@ group 'bio.terra'
 version '0.0.16-SNAPSHOT'
 sourceCompatibility = JavaVersion.VERSION_1_8
 
+def OPENCENSUS_VERSION = '0.28.2'
```
wchamber: Isaac had some gradle dependency management plugin for syncing versions across multiple libraries. https://github.com/DataBiosphere/terra-workspace-manager/blob/1f4038c4fb69834a3669b054a62db1b284cf26cb/build.gradle#L25-L32
I really don't know what's good practice here, but I hope you're making an informed decision :)

src/main/java/bio/terra/common/logging/LoggingInitializer.java
```diff
@@ -0,0 +1,52 @@
...
+      // propertySources.addLast(
+      //     new MapPropertySource(
+      //         "logging-properties", Map.of("terra.common.logging.human-readable-format",
+      // "true")));
+      return;
+    }
+
+    // propertySources.addLast(
+    //     new MapPropertySource(
+    //         "logging-properties", Map.of("terra.common.logging.google-format", "true")));
```
wchamber: forgotten commented out code?

src/main/java/bio/terra/common/logging/RequestLoggingFilter.java
```diff
@@ -0,0 +1,187 @@
...
+    HttpHeaders headers = new ServletServerHttpRequest(request).getHeaders();
+    json.put("headers", headers);
+
+    if (request instanceof ContentCachingRequestWrapper) {
+      ContentCachingRequestWrapper wrapper = (ContentCachingRequestWrapper) request;
+      byte[] buf = wrapper.getContentAsByteArray();
+      int length = Math.min(buf.length, 10000);
+      try {
+        json.put("payload", new String(buf, 0, length, wrapper.getCharacterEncoding()));
+      } catch (UnsupportedEncodingException e) {
```
wchamber, line 128: worth logging the exception at debug?

src/main/java/bio/terra/common/logging/RequestLoggingFilter.java
```diff
@@ -0,0 +1,187 @@
...
+    HttpHeaders headers = new ServletServerHttpResponse(response).getHeaders();
+    json.put("headers", headers);
+
+    if (response instanceof ContentCachingResponseWrapper) {
+      ContentCachingResponseWrapper wrapper = (ContentCachingResponseWrapper) response;
+      byte[] buf = wrapper.getContentAsByteArray();
+      int length = Math.min(buf.length, 10000);
+      try {
+        json.put("payload", new String(buf, 0, length, wrapper.getCharacterEncoding()));
+      } catch (UnsupportedEncodingException e) {
```
wchamber, line 152: ditto

src/test/java/bio/terra/common/logging/LoggingTest.java
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * <p>Inspired by
+ * https://github.com/eugenp/tutorials/blob/master/spring-boot-modules/spring-boot-testing/src/test/java/com/baeldung/testloglevel/LogbackMultiProfileTestLogLevelIntegrationTest.java
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT, classes = LoggingTestApplication.class)
+@ContextConfiguration(
+    // This is the simplest way to trigger LoggingInitializer from within this test. See
+    // LoggingTestApplication
+    // for an example of how a real-world application would initialize the logging flow.
```
wchamber: nit: whitespacing

src/main/java/bio/terra/common/logging/RequestLoggingFilter.java
```diff
@@ -0,0 +1,187 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.server.ServletServerHttpRequest;
+import org.springframework.http.server.ServletServerHttpResponse;
+import org.springframework.web.util.ContentCachingRequestWrapper;
+import org.springframework.web.util.ContentCachingResponseWrapper;
+
+/**
+ * A Servlet filter that collects and logs structured information about a HTTP requset and response.
```
ddietterich: nit typo

src/test/java/bio/terra/common/logging/LoggingTest.java
```diff
@@ -0,0 +1,197 @@
...
+  public static class MockSpanFilter implements Filter {
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+        throws IOException, ServletException {
+      Scope scope =
+          Tracing.getTracer().spanBuilderWithExplicitParent("test-span", null).startScopedSpan();
+      requestSpanContext = Tracing.getTracer().getCurrentSpan().getContext();
+      chain.doFilter(request, response);
+      // Note: we purposefully don't close the scope here. For some reason, it doesn't seem possible
+      // to inject this filter into the servlet at high enough precedence to
```
wchamber: fragment

src/main/java/bio/terra/common/logging/RequestLoggingFilter.java
```diff
@@ -0,0 +1,187 @@
...
+    json.setFactory(new JacksonFactory());
+    json.put("status", response.getStatus());
+
+    HttpHeaders headers = new ServletServerHttpResponse(response).getHeaders();
+    json.put("headers", headers);
+
+    if (response instanceof ContentCachingResponseWrapper) {
+      ContentCachingResponseWrapper wrapper = (ContentCachingResponseWrapper) response;
+      byte[] buf = wrapper.getContentAsByteArray();
+      int length = Math.min(buf.length, 10000);
```
ddietterich: Maybe give the loose constant a name for code doc'n.

ddietterich (commented):
Made some minor comments

src/main/java/bio/terra/common/logging/LoggingUtils.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.common.logging;
+
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+public class LoggingUtils {
```
wchamber, line 22: can some of these classes be package private? It'd be helpful for services to only have access to the classes from this package that they need to do something with.

wchamber (commented):
Having this done once centrally seems like a big win. But it doesn't look easy and I sure wouldn't know if you had something wrong :)

For open questions:
1. We started down the path of logging JSON objects (as strings) in logger calls in CRL. We try to serialize a json object to represent each request. The idea being it might be possible to do structured log searches for certain kinds of cloud requests. We haven't tried to utilize this, but you can see the appeal. See https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/5e5634c7f3b829ecbc6c9cb936ec5578da6a4c32/common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java#L131-L145

2. SGTM. Again with CRL, we thought that there might be services that want different levels of logging via logback config on different COWs (by means of their [class loggers through to the OperationAnnotator](https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/5e5634c7f3b829ecbc6c9cb936ec5578da6a4c32/google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java#L24-L37)  , but I don't think what's here precludes that?

build.gradle
gjuggler: Yeah – we should adopt that same practice here, I split out https://broadworkbench.atlassian.net/browse/PF-448 to track some Gradle housekeeping tasks.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
gjuggler: Thanks for commenting. The original StackdriverJsonLayout used GSON, and I wasn't 100% sure why. I thought maybe there's something to be said for using Google's JSON formatter for JSON that's ultimately going to be read by Google?

This prompted me to see what it took to switch to Jackson, and it was pretty straightforward. The LoggingUtils method needed some tiny changes, that's all.

Much better for us to be internally consistent.

src/main/java/bio/terra/common/logging/LoggingConfig.java
gjuggler: Good point – added some explicit notes.

src/main/java/bio/terra/common/logging/LoggingInitializer.java
gjuggler: Sorry about the commented-out code! I'd been toying with something more complicated that ultimately wasn't needed, I just forgot to remove it.

It's possible to base this off of a single property; I was just reflecting the interface used by WSM for this, see [docs](https://github.com/DataBiosphere/terra-workspace-manager/blob/38bef707e64be4b7c827516625f7d38c414ffb2c/DEVELOPMENT.md#tips) and [logback XML](https://github.com/DataBiosphere/terra-workspace-manager/blob/8740a0f7836039a6862455e7b02e75f6c1826ef2/src/main/resources/logback-spring.xml).

I don't have a strong opinion here – if it feels like a property is more straightforward, I'm happy to switch.

src/main/java/bio/terra/common/logging/LoggingUtils.java
gjuggler: Good idea – done where possible (I had to keep one public for mocking, but added @VisibleForTesting for clarity.

src/main/java/bio/terra/common/logging/RequestIdFilter.java
gjuggler: I'm porting code from WSM – presumably that passed code review at some point, and someone thought it was necessary. I'll plan to loop Isaac in for comment once this is past draft mode.

src/main/java/bio/terra/common/logging/RequestLoggingFilter.java
gjuggler: Good point – done, here and elsewhere.

src/main/java/bio/terra/common/logging/RequestLoggingFilter.java
gjuggler: Done

gjuggler (commented):
Thanks both for the quick look!

> 1. We started down the path of logging JSON objects (as strings) in logger calls in CRL. We try to serialize a json object to represent each request. The idea being it might be possible to do structured log searches for certain kinds of cloud requests. We haven't tried to utilize this, but you can see the appeal. See https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/5e5634c7f3b829ecbc6c9cb936ec5578da6a4c32/common/src/main/java/bio/terra/cloudres/common/OperationAnnotator.java#L131-L145

Thanks for the cross-link to CRL's approach. This would be very low-hanging fruit for structured logging. I added a new method to LoggingUtils which is more squarely targeted at that use case.

> 2. SGTM. Again with CRL, we thought that there might be services that want different levels of logging via logback config on different COWs (by means of their [class loggers through to the OperationAnnotator](https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/5e5634c7f3b829ecbc6c9cb936ec5578da6a4c32/google-bigquery/src/main/java/bio/terra/cloudres/google/bigquery/BigQueryCow.java#L24-L37)  , but I don't think what's here precludes that?

I don't think this should conflict with configuring per-package log levels, but I'll need to verify that before we incorporate into any downstream repos.

ddietterich:
Do you plan to implement the WM side of this (or some other service) to validate its construction?

gjuggler:
> Do you plan to implement the WM side of this (or some other service) to validate its construction?

Yes – I've been testing this primarily in local WSM as well as my personal env. (This is why I've been battling against personal env issues this week. It's also where the trace screenshot comes from.)

I polished up the loose ends there and put out https://github.com/DataBiosphere/terra-workspace-manager/pull/224 as a draft.

gjuggler:
OK, this feels developed enough to pull out of draft mode. Also adding @zarsky-broad to the reviewer list, as I think you were responsible for much of the request-id stuff in WSM that I'm extracting into this package.

wchamber (approved):

build.gradle
```diff
@@ -25,29 +25,43 @@ group 'bio.terra'
...
 dependencies {
     implementation group: 'bio.terra', name: 'stairway', version: '0.0.37-SNAPSHOT'
+    implementation group: 'ch.qos.logback.contrib', name: 'logback-json-classic', version: '0.1.5'
+    implementation group: 'ch.qos.logback.contrib', name: 'logback-jackson', version: '0.1.5'
+    implementation group: 'com.google.cloud', name: 'google-cloud-core', version: '1.94.0'
+    implementation group: 'com.google.apis', name: 'google-api-services-logging', version: 'v2-rev662-1.18.0-rc'
     implementation group: 'io.kubernetes', name: 'client-java', version: '10.0.0'
-    implementation group: "io.opencensus", name: "opencensus-api", version: "0.28.2"
+    implementation group: 'io.opencensus', name: 'opencensus-api', version: OPENCENSUS_VERSION
     implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.11'
```
zarsky-broad: version controlled by spring:

```suggestion
    implementation group: 'org.apache.commons', name: 'commons-lang3'
```

build.gradle
```diff
@@ -25,29 +25,43 @@ group 'bio.terra'
...
-    implementation group: 'org.slf4j', name: 'slf4j-api', version: '1.7.28'
-    implementation group: 'com.google.code.gson', name: 'gson', version: '2.8.6'
+    implementation group: 'org.hashids', name: 'hashids', version: '1.0.3'
     implementation group: 'org.liquibase' , name: 'liquibase-core', version: '3.10.0'
-    implementation group: 'org.springframework' , name: 'spring-context', version: '5.3.2'
+    implementation group: 'org.slf4j', name: 'slf4j-api', version: '1.7.28'
+    implementation group: 'org.springframework' , name: 'spring-context', version: '5.2.7.RELEASE'
+    implementation group: 'org.springframework' , name: 'spring-core', version: '5.2.7.RELEASE'
```
zarsky-broad: versions controlled by spring:
```suggestion
    implementation group: 'org.springframework' , name: 'spring-context'
    implementation group: 'org.springframework' , name: 'spring-core'
```

build.gradle
```diff
@@ -25,29 +25,43 @@ group 'bio.terra'
...
     api group: 'com.google.guava', name: 'guava', version: '29.0-jre'
 
     // For SpotBugs annotations (still FindBugs annotations)
     implementation 'com.google.code.findbugs:annotations:3.0.1'
 
     // For test
+    testImplementation group: 'com.jayway.jsonpath', name: 'json-path', version: '2.5.0'
+    testImplementation group: 'org.hamcrest', name: 'hamcrest', version: '2.2'
     testImplementation group: 'org.apache.commons', name: 'commons-dbcp2', version: '2.7.0'
     testImplementation group: "io.opencensus", name: "opencensus-impl", version: "0.28.2"
```
zarsky-broad:
```suggestion
    testImplementation group: 'io.opencensus', name: 'opencensus-impl', version: OPENCENSUS_VERSION
```

build.gradle
```diff
@@ -25,29 +25,43 @@ group 'bio.terra'
...
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: '5.6.2'
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: '5.6.2'
-    testImplementation group: 'org.hamcrest', name: 'hamcrest', version: '2.2'
     testImplementation group: 'org.mockito', name: 'mockito-core', version: '2.+'
+    testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-test'
+
+    annotationProcessor group: 'org.springframework.boot', name: 'spring-boot-configuration-processor'
 }
 
 dependencyLocking {
```
zarsky-broad, line 66: Looks like a lockfile was never actually created, run `./gradlew dependencies --write-locks`

build.gradle
zarsky-broad: The nifty part is the spring version management library is actually already included for spring itself, that's why you don't specify versions for e.g. the spring boot starters.

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -22,6 +18,9 @@
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
zarsky-broad: I get a lot of thrash if I run the intellij 'optimize imports' action, for instance it reverts this change. I'm just using the default google-java-format.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -0,0 +1,263 @@
...
+/**
+ * A Logback JSON layout aimed at generating log lines with useful details for ingestion by the
+ * Google Cloud Logging system.
+ *
+ * <p>See the LoggingTest integration test for more details on the output structure, including
+ * concrete examples of field values.
+ *
+ * <p>The output JSON relies heavily on Google Cloud's interpretation of structured JSON logs,
+ * including so-called "special fields in JSON payloads". See
+ * https://cloud.google.com/logging/docs/structured-logging#special-payload-fields for more details.
```
zarsky-broad: Make this a live link? (same request for other links in javadocs, but I won't repeat it)
```suggestion
 * including so-called "special fields in JSON payloads". See <a
 * href="https://cloud.google.com/logging/docs/structured-logging#special-payload-fields">Google's
 * documentation</a> for more details.
```

src/main/java/bio/terra/common/logging/RequestIdFilter.java
zarsky-broad: @ansingh7115 Any knowledge about this? I don't _think_ it was mine?

src/main/java/bio/terra/common/logging/RequestLoggingFilter.java
```diff
@@ -0,0 +1,216 @@
...
+      try {
+        map.put("payload", new String(buf, 0, length, wrapper.getCharacterEncoding()));
+      } catch (UnsupportedEncodingException e) {
+        log.debug("Error reading request payload", e);
+      }
+    }
+
+    log.debug(
+        "REQUEST: \n"
+            + new ObjectMapper().writerWithDefaultPrettyPrinter().writeValueAsString(map));
```
zarsky-broad:
```suggestion
        "REQUEST: \n{}",
        new ObjectMapper().writerWithDefaultPrettyPrinter().writeValueAsString(map));
```

src/main/java/bio/terra/common/logging/RequestLoggingFilter.java
```diff
@@ -0,0 +1,216 @@
...
+      int length = Math.min(buf.length, MAX_PAYLOAD_TO_DEBUG_LOG);
+      try {
+        map.put("payload", new String(buf, 0, length, wrapper.getCharacterEncoding()));
+      } catch (UnsupportedEncodingException e) {
+        log.debug("Error reading response payload", e);
+      }
+    }
+    log.debug(
+        "RESPONSE: \n"
+            + new ObjectMapper().writerWithDefaultPrettyPrinter().writeValueAsString(map));
```
zarsky-broad:
```suggestion
        "RESPONSE: \n{}",
        new ObjectMapper().writerWithDefaultPrettyPrinter().writeValueAsString(map));
```

zarsky-broad (commented):
some comments on consistency

build.gradle
gjuggler: Done. Thanks!

src/main/java/bio/terra/common/kubernetes/KubeService.java
gjuggler: I think spotless auto-formats things this way. I didn't make this change manually, and common-lib is set up to run spotless apply whenever Java is compiled.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
gjuggler: I don't have super strong feelings about this — as long as it's easy enough to click live links from within IntelliJ, since I almost never read Javadoc except from there.

src/main/java/bio/terra/common/kubernetes/KubeService.java
ddietterich: IntelliJ and spotless have different opinions about import ordering. Spotless wins, but it does mean the odd change. I would expect that it would resolve after awhile. If we always `spotless apply` on a build, everything we have in github should be in spotless format and we wouldn't get spurious diffs. Right?

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
ddietterich: I intend to use Stairway repo to build a generate GHA to post javadocs on github pages. That should make them more useful and worth debugging (and adding live links...)

zarsky-broad: So in intellij, without wrapping in an explicit `<a>`, the link is command clickable, but only if the documentation is displayed raw. If it's displaying the nicely formatted version, the `<a>` version is clickable, and the regular version isn't. FWIW I also only view it in intellij, mine just defaults to the nicely formatted version of javadoc comments, there should be a little toggle button in the margin.

src/main/java/bio/terra/common/kubernetes/KubeService.java
gjuggler: Occasionally I forget to build before committing changes and pushing my branch!

On the AoU side, we actually do a spotlessDiff as part of our unit tests, so a PR can't merge unless the format is exactly how Spotless would want it. Something we can consider if this becomes a persistent pain point.

build.gradle
```diff
@@ -130,6 +144,8 @@ task unitTest(type: Test) {
     }
 }
 
+// TODO(gjordan): turn off doclint to remove "no @return value" warnings
+// https://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html
```
ddietterich, line 148: Or fix the javadoc :)

src/main/java/bio/terra/common/logging/RequestIdFilter.java
ddietterich: I flagged this in my review of WSM as a question. What are the requirements of the id?
Would it be better as a ShortUUID so it would be guaranteed unique?

Anyway, it shouldn't stop merging this, but I would like to know. I'm happy to swing by and fix it when I do the UUID work.

ddietterich (approved):
LGtM

build.gradle
gjuggler: A valid option – but unnecessary return tags *feel* like noise to me, and I think Google's Java style guide agrees, since both their examples omit the return block ([ref](https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line)).

The fact that it's just a warning, and not an error, suggests Javadoc feels at least somewhat similarly :)

gjuggler:
> I flagged this in my review of WSM as a question. What are the requirements of the id?
> Would it be better as a ShortUUID so it would be guaranteed unique?

> Anyway, it shouldn't stop merging this, but I would like to know. I'm happy to swing by and fix it when I do the UUID work.

FWIW, I just checked AoU's implementation of this, and they used UUIDs. That's probably a reasonable default, unless we have some requirement that these are more human- or UI-friendly.

PR #22, opened 2021-02-23
[PF-283] Initial work on bio.terra.common.iam package.

jmczerk:
[Janitor](https://github.com/DataBiosphere/terra-resource-janitor/tree/master/src/main/java/bio/terra/janitor/service/iam), [TDR](https://github.com/DataBiosphere/jade-data-repo/tree/develop/src/main/java/bio/terra/service/iam), and [WSM](https://github.com/DataBiosphere/terra-workspace-manager/tree/dev/src/main/java/bio/terra/workspace/service/iam) have similar `UserRequestFactory` and `UserRequest` implementations.

This PR creates a new common `iam` package, with these classes being the lowest hanging fruit to generalize.

@wchamber raised some concerns around this breaking on-disk format.  The concern was originally driven by the change of `Optional<String>` to `String` in API's (which was removed due to adding no value after implementing a builder).  However, the [behavior of Jackson with `Optional<String>`](https://www.baeldung.com/jackson-optional) was manually confirmed to produce the same output as with a `String`, as long as we use the `Jdk8Module` with our` ObjectMapper`, [which we do](https://github.com/DataBiosphere/stairway/blob/develop/src/main/java/bio/terra/stairway/StairwayMapper.java#L26).

However, in my manual inspection to discover this, I realized that there may be a different issue that will break on-disk compatibility due to use of `enableDefaultTyping`, which adds the fully qualified class name to the serialized class:

`["bio.terra.common.iam.AuthenticatedUserRequest",{"email":"test@example.com","reqId":"78f84562-c442-49be-951a-a0a56230c35f","subjectId":"Subject","token":"0123.456-789AbCd"}]`

It seems like this will lead to a breakage when deserializing for systems with existing on-disk data that serialize as their local impl (i.e. `bio.terra.workspace.service.iamAuthenticatedUserRequest`) and switch to the common impl.  Thus the downstream PR's to consume this change will have upgrade considerations.

This might present a good opportunity to:
* Codify which fields should truly be optional here.
* Consider if we need the `reqId` UUID, as it doesn't seem to be consumed outside of tests.
* Consider if there's an opportunity to improve backward compatibility when serializing this class, as this required quite a bit of human intervention via manual inspection, and I'm still not 100% confident of the blast radius.


src/main/java/bio/terra/common/iam/AuthHeaderKeys.java
```diff
@@ -0,0 +1,19 @@
+package bio.terra.common.iam;
+
+/** Keys for auth headers */
+public enum AuthHeaderKeys {
+  OIDC_ACCESS_TOKEN("OIDC_ACCESS_token"),
```
ddietterich: Could these just be static final strings? It seems they are only used as a way to get strings.

Even if you leave it an enum, could it be moved inside of `ProxiedAuthenticatedUserRequestFactory`? Looks like that is the only place they are used.

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
```diff
@@ -0,0 +1,75 @@
...
+public class AuthenticatedUserRequestTest {
+
+  private static final Logger logger = LoggerFactory.getLogger(AuthenticatedUserRequestTest.class);
+
+  private static final String EmailAddress = "test@example.com";
+  private static final UUID RequestId = UUID.randomUUID();
+  private static final String SubjectId = "Subject";
+  private static final String Token = "0123.456-789AbCd";
+
+  public void validateJsonSerialization(AuthenticatedUserRequest request)
```
wchamber: nit: doesn't have to be public

ddietterich (approved):
One suggestion, otherwise LGTM

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
```diff
@@ -0,0 +1,75 @@
...
+      throws JsonProcessingException {
+    ObjectMapper objectMapper = new ObjectMapper();
+    String str = objectMapper.writeValueAsString(request);
```
wchamber: adding a test against deserializing hardcoded string is a good way to check for changes to backwards compatibility. While the test will be changed when the serialization format changes, it's a good reminder when the test breaks backwards compatibility and some assurance that it otherwise does not.

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
```diff
@@ -0,0 +1,75 @@
...
+    assertNull(req.getToken());
+    assertNotNull(req.getReqId());
+    assertThrows(ApiException.class, req::getRequiredToken);
+    validateJsonSerialization(req);
+  }
+
+  @Test
+  public void equality() {
+    AuthenticatedUserRequest req = new AuthenticatedUserRequest.Builder().build();
+    assertEquals(req, req);
```
wchamber: probably more interesting to assert equality for a full build.

src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
```diff
@@ -0,0 +1,110 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import java.util.Objects;
+import java.util.UUID;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+
+@JsonDeserialize(builder = AuthenticatedUserRequest.Builder.class)
+public class AuthenticatedUserRequest {
+
+  private final String email;
+  private final UUID reqId;
```
wchamber: if we're breaking backwards compatibility already, avoid abbreviations, use `requestId`. (if we can't remove this field entirely.) [ref](go/java-practices/naming#general-advice)

src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
```diff
@@ -0,0 +1,110 @@
...
+import bio.terra.common.exception.ApiException;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import java.util.Objects;
+import java.util.UUID;
+import org.apache.commons.lang.builder.HashCodeBuilder;
+
+@JsonDeserialize(builder = AuthenticatedUserRequest.Builder.class)
+public class AuthenticatedUserRequest {
```
wchamber, line 11: nit: all public classes should have a javadoc comment that's hopefully enlightening.
https://google.github.io/styleguide/javaguide.html#s7.3-javadoc-where-required

src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
```diff
@@ -0,0 +1,110 @@
...
+
+  private final String email;
+  private final UUID reqId;
+  private final String subjectId;
+  private final String token;
```
wchamber, line 15: nit: I could use comments for what both of these fields are.

wchamber (commented):
few nits (which I realize you've inherited, but no time like the present to make it better.)

Couple things for discussion with other reviewers:
1. My understanding of the `AuthenticatedUserRequest` is that it is not valid if any of email, subjectId, or token are null. Downstreams users of the class don't and shouldn't have to do null checks. We should throw an exception if we try to build the `AuthenticatedUserRequest` without them.
2. I think we should remove `reqId`. Having a way to associate a unique id with every request is a different concern from authenticating a user. It's also never used today AFAICT.
3. It's very difficult to not break backwards compatibility with things stored in Stairway FlightMaps, and it's easy to do without realizing it. The impact is that any flights in progress when the upgrade rolls out will not be able to be resumed. I think this is a general problem beyond the scope of this PR. We could add an explicit function for clients to use once to convert from the old style to the new style, and remove that after all services have been upgraded.

src/main/java/bio/terra/common/iam/AuthHeaderKeys.java
jmczerk: Done...  

This was a copy/paste, so I wasn't sure what value the Enum was adding over a String constant.  I also assume there was a thought that this might be reusable, but I also don't see anyone else using it...  We can always move it out in the future if there's a need to broaden the scope.

On a related note, I did want to move [LocalAuthenticatedUserRequestFactory](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/iam/LocalAuthenticatedUserRequestFactory.java) over from TDR as well, but balked due to the dependence on `ApplicationConfiguration`.  I would like to at least unify the parsing of the bearer token to use the regex logic I've added here for consistency.  What would be the idiomatic way to make this helper logic usable?

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
jmczerk: Done

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
jmczerk: I added a mechanism for easily adding hard-coded JSON test vectors.  If there's an easy way to grab some "real world" JSON examples from current stairway files, I would be happy to add those to the test vectors.

One thing to note is that it's also important that we're using the same Jackson `ObjectMapper` configuration as downstream serializers for this to be effective (there are some docs in the unit test on this).

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
jmczerk: Yes, this test was lacking in substance and documentation, I improved it in latest push.

src/main/java/bio/terra/common/iam/AuthHeaderKeys.java
ddietterich: The `Local...` version is only for testing. In TDR, the email and userid do not matter. I am not sure that is always the case. 

The idea is that in some test environment you are not running the front-end proxy that fills those fields in. So to have everything else work, you need a way to reliably fill in those fields.

wchamber: Without looking at the details, it might make sense to add the `LocalAuthenticatedUserRequestFactory` to a `testFixtures` part of the TCL. It'd be nice to share testing code as well, particularly as it's relevant to the common library. But we can certainly do that in a later PR.

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+    public void validate() throws JsonProcessingException {
+      AuthenticatedUserRequest deserialized =
+          objectMapper.readValue(this.json, AuthenticatedUserRequest.class);
+      assertEquals(request, deserialized);
+      assertEquals(request.hashCode(), deserialized.hashCode());
+    }
+  }
+
+  void validateJsonSerialization(AuthenticatedUserRequest request) throws JsonProcessingException {
```
wchamber: nit: this helper and the static class can both be private. Always make each class or member as inaccessible as possible, [ref](go/java-practices/visibility).

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
```diff
@@ -0,0 +1,158 @@
...
+  static class TestVector {
+    private final String json;
+    private final AuthenticatedUserRequest request;
+
+    public TestVector(String json, AuthenticatedUserRequest request) {
+      this.json = json;
+      this.request = request;
+    }
+
+    public void validate() throws JsonProcessingException {
```
wchamber: optional style: I'm not totally convinced this is worth having a class for. I think making the `validate` function into `assertDeserializedEquals(String json, AuthenticatedUserRequest expectedRequest)` (or some better function name) and inlining the `testVectors` would keep the indirection in tests to a minimum. But this is very much on the level of stylistic opinions, and there might be a well established pattern here I'm not familiar with :)

wchamber (approved):
Couple more nits, but lgtm. Let's discuss the backwards incompatible changes & rolling this out offline. I think it would be fine to start here and do more tweaks in a later PR as well.

src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
ddietterich: The fact of moving the code to a different package will break on-disk compatibility in data serialized in Stairway.

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
jmczerk: Yeah, in my mind I was thinking this would look a lot more compact than it turned out; also the list of test vectors will not be cumulative so there's not much to be gained by adding syntactic sugar even if it had panned out.  Simplified it.

src/main/java/bio/terra/common/iam/AuthHeaderKeys.java
jmczerk: Yeah, decoupling from the local app config will take a little extra work, I'll consider revisiting when I get to the consuming the changes in TDR.

src/main/java/bio/terra/common/iam/BearerTokenParser.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+    return out;
+  }
+
+  /**
+   * Takes the Authorization header passed as a String in {@code authorizationHeader}, parses out
+   * the bearer token payload, and calls the passed Consumer function with the payload passed as a
+   * String. If the header is not properly formatted, then the passed Consumer is not called.
+   */
+  public static void parseInto(String authorizationHeader, Consumer<String> op) {
```
wchamber: Empty string return value + parseInto seems a little funny. In general, special case return values are not great. I think it'd be clearer if we returned a `@Nullable String` or an `Optional<String>` from `parse`. Then I don't think we need to have `parseInto` at all and can handle failure to parse at `parse`'s call site. Having a special `Consumer` function for a null case may be overkill.

src/main/java/bio/terra/common/iam/BearerTokenParser.java
```diff
@@ -0,0 +1,44 @@
+package bio.terra.common.iam;
+
+import java.util.function.Consumer;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/** Utility class for consistently parsing bearer tokens from HTTP Request Authorization headers. */
+public final class BearerTokenParser {
```
wchamber, line 8: optional nit: I don't know that this is worth having as a whole separate class yet. It doesn't seem particularly likely to me that we're going to re-use it. Consider including the `parse` function back within the `Proxied...` class. We can mark `parse` as `@VisibleForTesting` if this is a testing concern.

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
```diff
@@ -44,48 +39,41 @@
...
+    assertThrows(IllegalStateException.class, builder::build);
+    builder.setEmail(EMAIL_ADDRESS);
+
+    // Build fails due to no Subject
+    assertThrows(IllegalStateException.class, builder::build);
+    builder.setSubjectId(SUBJECT_ID);
+
+    // Build fails due to no Token
+    assertThrows(IllegalStateException.class, builder::build);
+    builder.setToken(TOKEN);
```
wchamber, line 60: this test would pass even if it was allowed to not set the email or the subject given that the token is set last. Consider going `validRequest.toBuilder().setEmail(null).build()` for each field.

src/main/java/bio/terra/common/iam/ProxiedAuthenticatedUserRequestFactory.java
```diff
@@ -43,10 +29,7 @@ public AuthenticatedUserRequest from(HttpServletRequest servletRequest) {
...
       String authHeader = servletRequest.getHeader(AUTHORIZATION);
       if (authHeader != null) {
-        Matcher matcher = authorizationPattern.matcher(authHeader);
-        if (matcher.matches()) {
-          builder.setToken(matcher.group("token"));
-        }
+        BearerTokenParser.parseInto(authHeader, builder::setToken);
       }
     }
     return builder.build();
```
wchamber: consider throwing a specific class of exception, like if we weren't able to build instead of the very generic `IllegalStateException`. `ProxyUnauthorizedException`? naming is hard... Should maybe subclass [UnauthorizedException](https://github.com/DataBiosphere/terra-common-lib/blob/develop/src/main/java/bio/terra/common/exception/UnauthorizedException.java)?

wchamber (approved):

src/main/java/bio/terra/common/iam/BearerTokenParser.java
jmczerk: This one was intentional, and related to the`LocalAuth...` class mentioned before.  Basically I saw three ways that this header was being parsed, all three different but equally fragile.  So there's at least one other place I expect to use it when available.

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
jmczerk: Note that I'm reusing the local instance `builder` for this very reason, I did verify the coverage using the IntelliJ run-with-coverage functionality.  On a related note, I'd love tho see [OpenClover](https://openclover.org/) coverage reports as part of our gradle build artifacts to make this less manual/interactive; I'll look into how to make that happen.

src/main/java/bio/terra/common/iam/BearerTokenParser.java
jmczerk: Yeah, I was kind of torn on what to return, thus the attempt to sidestep it with `parseInto`.  I'm really more inclined to throw in `parse` if the regex match fails, as there's probably no good reason for having a scrozzled Auth header.

src/main/java/bio/terra/common/iam/ProxiedAuthenticatedUserRequestFactory.java
jmczerk: Makes sense, would you recommend still having `build()` throw `IllegalStateExcpetion` per the [Java Practices Doc](https://engdoc.corp.google.com/eng/doc/devguide/java/practices/builders.md?cl=head#api), and try/catch/throw an auth specific exception here (aligned with the discussion about throwing in the bearer token parser utility method)?

jmczerk:
@wchamber I think I've addressed the outstanding concerns.  If you can give the most recent commit (not the fixup commit) a once over, I will merge this if no show-stoppers.

src/main/java/bio/terra/common/iam/ProxiedAuthenticatedUserRequestFactory.java
wchamber: I think `build()` throwing an IllegalStateException still makes sense (the builder doesn't know it's getting bad values because the user isn't authenticated, it's just not getting what it needds). I think catching the IllegalStateException and rethrowing it as the cause of an auth specific exception would work fine. Or throwing an auth exception before passing any null values to the builder. But it's also ok to leave it as throwing unauthorized on build.

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
```diff
@@ -0,0 +1,140 @@
...
+    assertNotEquals(req, req.toBuilder().setToken("JUNK").build());
+
+    // Explicit test for comparison to self
+    assertEquals(req, req);
+
+    // Explicit test for off-type comparison
+    assertNotEquals(req, "test");
+  }
+
+  static class TestVector {
```
wchamber: I don't think this class is being used anymore?

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
wchamber: I think having tests that cover all paths would be better than relying on code coverage tooling, but not a big deal. Easier to ignore a code coverage warning than a failing test.

wchamber (approved):

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
jmczerk: Hmmm... not sure I follow, as this is covering all paths due to multiple calls on the same builder isntance (created on line 48).  Lets discuss offline.

PR #25, opened 2021-03-02
[PF-511] Logging package: fix unit test and spotbugs errors

gjuggler:
Fixes a couple bugs in the verification / static checks for the logging package.

I hadn't fully tested the `LoggingUtils` init method before merging #19. Both unit test classes work fine when run independently, but either one fails (whichever one is run second) when run together as part of a single JUnit execution.

The problem was with the `loggingInitialized` boolean. It works fine when only one app initialization flow is occurring within the lifetime of the Java execution, but in the test environment a new app context is initialized for each test class.

The cleanest fix was to add some code which explicitly resets the Logback context and reinstates human-readable logging, to complement the existing code which resets Logback and reinstates JSON logging.

This PR also contains a drive-by cleanup, to make the logging tests use the newer JUnit Jupiter test APIs rather than legacy JUnit 4 equivalents.


wchamber (approved):

src/main/java/bio/terra/common/logging/LoggingUtils.java
```diff
@@ -78,34 +77,30 @@ static JsonNode jsonFromString(String s) throws JsonProcessingException {
...
-
     ch.qos.logback.classic.Logger logbackLogger =
         (ch.qos.logback.classic.Logger) LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
     ConfigurableEnvironment environment = applicationContext.getEnvironment();
 
     if (Arrays.stream(environment.getActiveProfiles()).anyMatch("human-readable-logging"::equals)) {
       System.out.println("Human-readable logging enabled, skipping Google JSON layout");
-      return;
-    }
+      // No action needed. The logback.xml file on the classpath contains a default human-readable
```
jaycarlton, line 86: curious: If this branch is effectively the default behavior, why was it the one to get the keyword?

src/main/java/bio/terra/common/logging/LoggingUtils.java
```diff
@@ -78,34 +77,30 @@ static JsonNode jsonFromString(String s) throws JsonProcessingException {
...
+      encoder.start();
 
-    LayoutWrappingEncoder encoder = new LayoutWrappingEncoder();
-    encoder.setLayout(layout);
-    encoder.start();
+      ConsoleAppender appender = new ConsoleAppender();
+      appender.setName(TERRA_APPENDER_NAME);
+      appender.setEncoder(encoder);
+      appender.setContext(logbackLogger.getLoggerContext());
+      appender.start();
```
jaycarlton, line 100: 🎵 the ApplicationContext on the layout and the layout on the encoder and the encoder on the appender and the appender on the logger and the green grass grows all around all around...

src/test/java/bio/terra/common/logging/LoggingTest.java
```diff
@@ -81,8 +77,7 @@
     @Override
     public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
         throws IOException, ServletException {
-      Scope scope =
-          Tracing.getTracer().spanBuilderWithExplicitParent("test-span", null).startScopedSpan();
+      Tracing.getTracer().spanBuilderWithExplicitParent("test-span", null).startScopedSpan();
       requestSpanContext = Tracing.getTracer().getCurrentSpan().getContext();
       chain.doFilter(request, response);
       // Note: we purposefully don't close the scope here. For some reason, it doesn't seem possible
```
jaycarlton, line 83: This sounds odd. Possibly out of scope.

src/test/java/bio/terra/common/logging/LoggingTest.java
```diff
@@ -201,8 +196,8 @@ public void testStructuredLogging() {
             .read(path);
   }
 
-  private String lastLoggedLine() {
-    String[] lines = outputCapture.getAll().split("\n");
+  private String lastLoggedLine(CapturedOutput capturedOutput) {
```
jaycarlton, line 199: should probably be Optional<String> in case there are no lines in the capture.

jaycarlton (approved):

src/test/java/bio/terra/common/logging/LoggingTest.java
gjuggler: That might be a tad cleaner, but IMO it's good enough for the test to fail on an NPE here. If you're asking for the last logged line and none exists, the case should fail hard & fast.

src/test/java/bio/terra/common/logging/LoggingTest.java
gjuggler: :) it's a condensation of a much more frustrating debugging session than one might imagine...

PR #26, opened 2021-03-03
[PF-448] Gradle cleanups: adopt quality plugin and factor out snippets

gjuggler:
This PR makes a few changes to the Gradle setup for common-lib:
 - Incorporate https://github.com/xvik/gradle-quality-plugin, which provides a better wrapper around Spotbugs than the spotbugs-gradle-plugin and also supports execution of multiple tools. This is similar to https://github.com/DataBiosphere/terra-workspace-manager/pull/243.
 - Split out distinct snippets of Gradle config into separate files. This reduces the mental burden of scrolling through one large file, and gives service repos an easier path to cloning certain standard pieces of config. (This is basically a waypoint on the path towards producing a bona fide Gradle platform plugin... to do it "right" will take a dedicated effort, but IMO this has some benefits for code clarity in its own right.)
 - Remove the separate 'unitTest' Gradle task (we don't currently split out test suites for common-lib, it's simpler to leave them all in "test") and adjust the tasks called by the CI/CD workflows. The "gradle build" command takes care of everything except javadoc, which is on a separate dependency thread. See the diagram of dependencies [here](https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_tasks).

(N.B. This PR is not the primary Gradle cleanup mentioned in PF-448, but it's spiritually in line with that ticket, so I tagged it there.)


src/main/java/bio/terra/common/kubernetes/KubePodListener.java
```diff
@@ -80,6 +80,7 @@
   }
 
   @Override
+  @SuppressWarnings({"CyclomaticComplexity", "ExecutableStatementCount", "NPathComplexity"})
```
wchamber: I agree this method does too much, but I also don't really want to have to deal with stylistic warnings. Why not exclude these as well? I think this is a problem for code review, not a problem for automated analysis.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -55,6 +55,7 @@
  * href="https://github.com/ankurcha/gcloud-logging-slf4j-logback/">gcloud-logging-slf4j-logback</a>
  * which inspired some of the patterns used here.
  */
+@SuppressWarnings("MultipleStringLiterals")
```
wchamber: ditto I don't think good style is automatable and adding suppression warnings feels silly.

src/main/java/bio/terra/common/stairway/TracingHook.java
```diff
@@ -174,6 +174,7 @@ public HookAction end(FlightContext flightContext) {
    * Gets or creates the context of the submission Span. Each Flight span is linked to the
    * submission span to correlate all Flight Spans.
    */
+  @SuppressWarnings("ReturnCount")
```
wchamber: IMHO, early return is a good pattern, not a bad one. go/tott/455

wchamber (commented):
How did you come up with the exclusions & inclusions that you did?
I'm very suspicious of trying to delegate style to static code analysis. I'd like to have more of a discussion on where we draw the line and what would be a good/bad reason to add `SuppressWarning`.

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
jaycarlton: I'd file a ticket to look at the method if it's tripping these complexity thresholds, or at least comment why it must be this complex/all one method. These metrics are nice because of how formulaic they are, and they've been around forever.

build.gradle
```diff
@@ -3,12 +3,13 @@ plugins {
     id 'java-library'
     id 'maven-publish'
 
+    id 'com.diffplug.spotless' version '5.10.2'
     id 'com.github.ben-manes.versions' version '0.33.0'
-    id 'com.github.spotbugs' version '4.5.1'
-    id 'com.diffplug.gradle.spotless' version '3.27.2'
     id 'com.jfrog.artifactory' version '4.13.0'
     id 'io.spring.dependency-management' version '1.0.9.RELEASE'
+    id 'jacoco'
```
jaycarlton: should this be up with the other non-versioned IDs? Should those be alphabetized as well?

build.gradle
```diff
@@ -88,94 +76,11 @@ if (hasProperty('buildScan')) {
...
-    withJavadocJar()
-}
-
-def artifactory_repo_key = System.getenv("ARTIFACTORY_REPO_KEY") != null ? System.getenv("ARTIFACTORY_REPO_KEY") : 'libs-snapshot-local'
-def artifactory_username = System.getenv('ARTIFACTORY_USERNAME')
-def artifactory_password = System.getenv('ARTIFACTORY_PASSWORD')
+apply from: "$rootDir/gradle/dependency-locking.gradle"
+apply from: "$rootDir/gradle/javadoc.gradle"
+apply from: "$rootDir/gradle/publishing.gradle"
+apply from: "$rootDir/gradle/quality.gradle"
```
jaycarlton, line 95: Is the idea that I'll be able to do an external apply from in the other repo and pick up this file? I'd still have duplicate xml configs.

gradle/config/checkstyle/suppressions.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
+<!DOCTYPE suppressions PUBLIC
+    "-//Checkstyle//DTD SuppressionFilter Configuration 1.2//EN"
+    "https://checkstyle.org/dtds/suppressions_1_2.dtd">
+
+<!-- Empty suppressions file: copy it into gradle/config/checkstyle and specify required suppressions
+     (https://checkstyle.sourceforge.io/config_filters.html#SuppressionFilter)-->
+<suppressions>
+
```
jaycarlton: I had to add `<suppress checks="[a-zA-Z0-9]*" files="[\\/]generated[\\/]"/>` to skip all  generated files.

src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
```diff
@@ -45,7 +45,7 @@ private DatabaseRetryUtils() {}
   }
 
   /**
-   * Tests an exception to see if it is retryable
+   * Tests an exception to see if it is retryable.
```
jaycarlton, line 48: I would almost suppress this one globally. I like to use fragments in comments, and the  absence of a trailing period can have some meaning.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
jaycarlton: for spotbugs at least, the new annotation is `@SuppressFBWarnings` http://findbugs.sourceforge.net/api/edu/umd/cs/findbugs/annotations/SuppressFBWarnings.html

src/main/java/bio/terra/common/stairway/TracingHook.java
jaycarlton: I'd adjust the max value here rather than suppress. More than about 4 is suspicious I'd think. Default is 2.

jaycarlton (approved):
It would be good to make a list of the global exceptions and run through them in a team setting, especially if they're not directly covered in the style guide. 

I'm not sure what to do with my own PRs yet. I think I'll wait for this one and then synchronize the config files.

jaycarlton:
We also have the option of turning on the tools one at a time. Starting with spotless only would be the easiest way. Then the PRs to turn them on could come with lists of exclusions for individual tools, and we'd get a better sense of which tools were good for what.

gjuggler:
> How did you come up with the exclusions & inclusions that you did?
> I'm very suspicious of trying to delegate style to static code analysis. I'd like to have more of a discussion on where we draw the line and what would be a good/bad reason to add `SuppressWarning`.

Good points. I tried to draft some guiding principles here. See docs in gradle/quality.gradle – feedback or thoughts?

There is no doubt that fine-tuning something like this will be an ongoing process. I spent maybe 20 minutes adding exclusions for the most redundant / obvious false positives.

I think the overall opportunity is sound, though; already the findings catch some pretty low-effort and helpful fixes (see https://gist.github.com/gjuggler/b9c32894ad2f932a9c967589da7281ad).

gradle/quality.gradle
```diff
@@ -0,0 +1,31 @@
...
+ * Terra configs for the Gradle quality plugin.
+ *
+ * See docs at http://xvik.github.io/gradle-quality-plugin/4.5.0/
+ *
+ * Guiding principles behind our use of static code analyzers:
+ *
+ * 1) Shift left. We value code analyzers because they can accelerate and
+ *    automate our detection of clear-cut violations of Java style expectations
+ *    and good coding practice. Automated enforcement should be scoped to
+ *    non-controversial matters of code health.
```
wchamber, line 11: Automated enforcement should *only* be scoped to...
my 2c add 
```
2) Automated enforcement is for programming errors or mechanical code formatting. Style and good judgment can not be replaced by automatic checks. Good candidates for automated enforcement include code formatting, such as white space or camel case variable naming, or programmer errors, such as unused or uninitialized variables. Bad candidates for automated enforcement are include trying to determine when a method is too complex or coding conventions without such as class member ordering.
```
and follow up with `3) Suppress early...`

gradle/config/pmd/pmd.xml
```diff
@@ -0,0 +1,167 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ruleset name="Base ruleset" xmlns="http://pmd.sourceforge.net/ruleset/2.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://pmd.sourceforge.net/ruleset/2.0.0 http://pmd.sourceforge.net/ruleset_2_0_0.xsd">
+
+    <description>
+        General Java quality rules.
```
wchamber, line 7: Some exclusion suggestions:

[Whitespace | WhitespaceAround] bio.terra.common.db.(DatabaseRetryUtils.java:16)
  '{' is not followed by whitespace.
  http://checkstyle.sourceforge.io/config_whitespace.html#WhitespaceAround
looks like there's some ways to limit it, to allow various types of things, but doesn't seem particularly useful to hit every reason to have whitespace around brackets.

style-ish things:
[Coding | DeclarationOrder] bio.terra.common.iam.(AuthenticatedUserRequest.java:91)
  Constructor definition in wrong order.
  http://checkstyle.sourceforge.io/config_coding.html#DeclarationOrder
[Metrics | JavaNCSS] bio.terra.common.kubernetes.(KubePodListener.java:82)
  NCSS for this method is 51 (max allowed is 50).
  http://checkstyle.sourceforge.io/config_metrics.html#JavaNCSS
​
[Coding | ReturnCount] bio.terra.common.kubernetes.(KubePodListener.java:82)
  Return count is 4 (max allowed for void methods/constructors/lambdas is 1).
  http://checkstyle.sourceforge.io/config_coding.html#ReturnCount
[Coding | ReturnCount] bio.terra.common.logging.(GoogleJsonLayout.java:220)
  Return count is 2 (max allowed for void methods/constructors/lambdas is 1).
  http://checkstyle.sourceforge.io/config_coding.html#ReturnCount
[Metrics | ClassFanOutComplexity] bio.terra.common.logging.(RequestLoggingFilter.java:45)
  Class Fan-Out Complexity is 22 (max allowed is 20).
  http://checkstyle.sourceforge.io/config_metrics.html#ClassFanOutComplexity

premature optimization suggestions
[Performance | InefficientStringBuffering] bio.terra.common.exception.(AbstractGlobalExceptionHandler.java:61)
  Avoid concatenating nonliterals in a StringBuffer/StringBuilder constructor or append().
  https://pmd.github.io/pmd-6.31.0/pmd_rules_java_performance.html#inefficientstringbuffering

[Code Style | LinguisticNaming] bio.terra.common.iam.(AuthenticatedUserRequest.java:46)
  Linguistics Antipattern - The setter 'setEmail' should not return any type except void linguistically
  https://pmd.github.io/pmd-6.31.0/pmd_rules_java_codestyle.html#linguisticnaming
[Design | NcssCount] bio.terra.common.kubernetes.(KubePodListener.java:84)
  The method 'run()' has a NCSS line count of 52.
  https://pmd.github.io/pmd-6.31.0/pmd_rules_java_design.html#ncsscount

[Design | CyclomaticComplexity] bio.terra.common.kubernetes.(KubePodListener.java:84)
  The method 'run()' has a cyclomatic complexity of 14.
  https://pmd.github.io/pmd-6.31.0/pmd_rules_java_design.html#cyclomaticcomplexity
​
[Error Prone | AvoidFieldNameMatchingMethodName] bio.terra.common.kubernetes.(KubeService.java:49)
  Field isShutdown has the same name as a method
  https://pmd.github.io/pmd-6.31.0/pmd_rules_java_errorprone.html#avoidfieldnamematchingmethodname

gradle/quality.gradle
wchamber: Consider Google's Error Prone criteria for checks: https://github.com/google/error-prone/wiki/Criteria-for-new-checks

build.gradle
gjuggler: The idea here is to make our main Gradle file less monolithic, with clearer separation of concerns in each sub-config. It's purely about local organization. It will encourage cleaner copy-pasting of Gradle config between repos for as long as we're in that paradigm. Longer term, it seems clear that providing a central Gradle plugin is the right way to share config like this.

gradle/config/pmd/pmd.xml
gjuggler: Done

gradle/quality.gradle
gjuggler: Thanks – added

src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
gjuggler: Done

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
gjuggler: In light of the principles we've drawn up, I'm opting to exclude these rules.

src/main/java/bio/terra/common/stairway/TracingHook.java
gjuggler: Similar story here; I've opted to exclude this rule.

gjuggler (commented):
Picking this one up briefly while I'm knee deep in some of the code (maybe for the last time, ha). I'll hold off from merging until Will has a chance for another look next week, but I think it should be in pretty good shape now.

Note that I clarified the name of the main Gradle command in the GHA workflow. As long as we include the `withJavadocJar()` statement in the javadoc.gradle config, a simple `./gradlew build` invocation should run all Spotless, Javadoc, and unit test checks that we need in order to ensure a clean build.

src/main/java/bio/terra/common/logging/RequestLoggingFilter.java
```diff
@@ -81,7 +80,7 @@ public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain)
         getGoogleHttpRequestObject(request, response, latency, requestSize, responseSize));
     logPayload.put("requestHeaders", getRequestHeaders(request));
 
-    String requestPath = null;
+    String requestPath;
```
jaycarlton, line 83: you can declare this `final` so the compiler keeps you honest and makes sure it's initialized.

gjuggler: I see `error: variable requestPath might already have been assigned` if I try that...

jaycarlton: Maybe catch uses the same scope as try so it's triggering that warning. Or does it not compile?

gjuggler: Yeah, I should have clarified it's a `javac` compilation error. Given this is not at all the purpose of this PR, I'm not going to press further.

gradle/config/pmd/pmd.xml
```diff
@@ -0,0 +1,202 @@
...
+    <rule ref="category/java/design.xml/AvoidDeeplyNestedIfStmts">
+        <properties>
+            <property name="problemDepth" value="4"/>
+        </properties>
+    </rule>
+    <!--
+      Terra: complexity measures are squishy and vague. Prefer to catch complexity in code review.
+
+      (The commented-out sections below were present in the default geneated XML. They are kept
+      here in commented form for clarity and provenance.)
```
wchamber, line 104: I don't get why they're commented out instead of deleted. Are we ever going to refer to the default generated XML again?

gradle/test.gradle
```diff
@@ -0,0 +1,55 @@
...
+    testImplementation group: 'com.jayway.jsonpath', name: 'json-path'
+    testImplementation group: 'io.opencensus', name: 'opencensus-impl', version: OPENCENSUS_VERSION
+    testImplementation group: 'org.apache.commons', name: 'commons-dbcp2', version: '2.7.0'
+    testImplementation group: 'org.hamcrest', name: 'hamcrest', version: '2.2'
+    // Jupiter version is managed by the Spring Boot plugin
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter'
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api'
+    testImplementation group: 'org.mockito', name: 'mockito-core', version: '2.+'
+    testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-test'
+    testImplementation group: 'org.openapitools', name: 'jackson-databind-nullable', version: '0.2.1'
```
wchamber: I'm not totally sold on keeping the test dependencies in a separate place. I'm worried about the need to look at two files for dependencies. It's possible for dependencies to move from test-only to non-test overtime. It would probably be bad if we had a different `opencensus-impl` or `jackson-databind-nullable` in tests vs normally, but splitting it into two files feels like it's harder to catch that kind of thing.
Could be persuaded.

src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
```diff
@@ -14,6 +14,12 @@
   private final String subjectId;
   private final String token;
 
+  protected AuthenticatedUserRequest(Builder builder) {
```
wchamber: nit: if we're going to move the constructor, let's make it `private` and even make the class final. `protected` tells me we want subclasses, but we don't want subclasses by default

src/main/java/bio/terra/common/stairway/TracingHook.java
```diff
@@ -92,7 +92,7 @@ public static void storeCurrentContextAsSubmission(FlightMap inputMap) {
    * <p>Note that the same Flight may be run multiple times by Stairway.
    */
   private class TraceFlightHook implements DynamicHook {
-    private Scope flightScope = null;
+    private Scope flightScope;
```
wchamber, line 95: no action required: go/java-practices/redundancy#default-initialization-for-non-final-fields :)

wchamber (approved):

gradle/test.gradle
gjuggler: SGTM, I was on the fence here, so your concern makes it an easy call to undo.

src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
gjuggler: Done

gjuggler: Making the constructor private works. Marking as final doesn't, due to some Jackson requirements from the unit tests (which AIUI accurately reflects the current default serialization in Stairway).

gradle/config/pmd/pmd.xml
gjuggler: Yeah, I understand the aversion to checking in commented-out stuff. But for this type of config, where we're taking something "stock" and making targeted changes particular to our environment, I think deletions are just as relevant as additions.

Imagine someone a year from now wants to upgrade to a new major version of the quality plugin. Keeping this here makes it a bit easier to understand what opinions we had re: specific suppressions and why, and re-apply the same changes if needed to a newly-generated stock config.

PR #27, opened 2021-03-05
[PF-541] refactor kube pod listener

jaycarlton:
I was studying this class and had a little trouble following what it does, so I broke it into some  smaller functions. Also added `final`, since some variables are imperative and some  aren't, some stream and Optional usage, and a couple of comments.

### DIFF
--- a/src/main/java/bio/terra/common/kubernetes/KubePodListener.java
+++ b/src/main/java/bio/terra/common/kubernetes/KubePodListener.java
@@ -92,19 +92,20 @@ class KubePodListener implements Runnable {
     // retry loop
     while (true) {
       try {
-        final ApiClient defaultClient = Config.defaultClient();
+        final ApiClient client = Config.defaultClient();
         // infinite timeout
         final OkHttpClient httpClient =
-            defaultClient.getHttpClient().newBuilder().readTimeout(0, TimeUnit.SECONDS).build();
-        defaultClient.setHttpClient(httpClient);
-        Configuration.setDefaultApiClient(defaultClient);
+            client.getHttpClient().newBuilder().readTimeout(0, TimeUnit.SECONDS).build();
+        client.setHttpClient(httpClient);
+        Configuration.setDefaultApiClient(client);
         final CoreV1Api kubeApi = new CoreV1Api();
 
         // Instantiate the watch and scope it.
-        try (Watch<V1Namespace> watch = makeWatch(defaultClient, kubeApi)) {
+        try (Watch<V1Namespace> watch = makeWatch(client, kubeApi)) {
           for (Watch.Response<V1Namespace> item : watch) {
             // If we are shutting down, we stop watching
             if (kubeService.isShutdown()) {
+              logger.info("Kubernetes service is shutting down. Exiting.");
               return;
             }
             // Reset retry if the watch worked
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/common/kubernetes/KubePodListener.java
```diff
@@ -111,34 +111,16 @@ public void run() {
...
-                  }
-                } catch (DatabaseOperationException | StairwayExecutionException ex) {
-                  logger.error("Stairway recoverStairway failed to recover pod: " + podName, ex);
-                } catch (InterruptedException ex) {
-                  logger.info("KubePodListener interrupted - exiting", ex);
-                  exception = ex;
-                  return;
-                }
-              }
+            if (handlePodOperation(operation, podName)) {
```
jaycarlton: This is the most awkward bit. It returns a boolean letting you know to return early, but the name almost needs to be `handlePodOperation_isInterrupted`. 🤢

jaycarlton:
Looks like there's no unit test coverage for this guy, so I'm thinking I should add some. The mocking might be wonky but I think it should work.

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
```diff
@@ -86,22 +88,20 @@ public void run() {
     exception = null;
     int consecutiveRetryCount = 0;
     int retryWait = WATCH_INITIAL_WAIT;
+
+    // retry loop
     while (true) {
       try {
-        ApiClient client = Config.defaultClient();
+        final ApiClient defaultClient = Config.defaultClient();
```
ddietterich: I prefer client. There is no other client, so calling it default doesn't inform anything.

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
ddietterich: How about
- make a void method for Add
- make a void method for Delete
- move the InterruptedException catch to the outer try and return from there.

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
```diff
@@ -111,34 +111,16 @@ public void run() {
...
-            String operation = item.type;
-            String podName =
-                (item.object.getMetadata() != null)
-                    ? item.object.getMetadata().getName()
-                    : org.apache.commons.lang3.StringUtils.EMPTY;
+            final String operation = item.type;
+            final String podName =
+                Optional.ofNullable(item.object.getMetadata())
+                    .map(V1ObjectMeta::getName)
+                    .orElse("");
```
ddietterich: What's wrong with StringUtils.empty?

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
```diff
@@ -177,14 +210,10 @@ Exception getException() {
...
-    int count = 0;
-    for (PodState podState : podMap.values()) {
-      if (podState == PodState.RUNNING) {
-        count++;
-      }
-    }
+    final long count = podMap.values().stream().filter(PodState.RUNNING::equals).count();
     logger.info("KubePodListener ActivePodCount: {} of {} pods active", count, podMap.size());
-    return count;
+    return (int) count;
```
ddietterich: Maybe use Math.toIntExact instead of the cast.

ddietterich (commented):
Thanks for the cleanup effort! A few comments and a suggested alternative.

ddietterich:
I don't think it is worth trying to unit test this. Seems to me a lot of effort for little benefit. The whole purpose of the code is to interact with K8s, so I believe in testing it with K8s.

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
jaycarlton: I'd never seen it before. Is there an issue with `""`? I thought that was standard.

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
jaycarlton: oh nice

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
jaycarlton: That works all right, thanks.

src/main/java/bio/terra/common/kubernetes/KubePodListener.java
ddietterich: It is probably bit bumming, but every "" creates a String object. Using EMPTY shares one.

ddietterich (approved):
I like it!

tnadler (approved):
Super helpful in making this file understandable, thank you!

PR #28, opened 2021-03-08
PF-279 Implement a common tracing package

wchamber:
Implement a common tracing package to be used by Terra services.

- Sets up exporting traces to Stackdriver
- Sets up automatic request tracing for http service requests with terra attributes like the requestId.
- Allows use of an `@Traced` annotation to add trace spans around bean methods.

This adds our first common library configuration properties under `terra.common.tracing` for controlling the sampling rate, what URLs to add spans for, and whether to enable Stackdriver exporting.

Note that this disables the io.opencensus.contrib.spring auto configuration that Workspace Manager currently relies on today to create traces for each request to the service. We should update WM to use this new package when we update it for this new version of the library. See https://github.com/DataBiosphere/terra-workspace-manager/pull/248

Example trace from my personal environment with WM using this package: https://console.cloud.google.com/traces/list?tid=e3423c85586219bc1ea6808f20dc7dd2


src/main/java/bio/terra/common/tracing/disable/AutoConfigDisable.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
+
+/**
+ * Disable the io.opencensus.contrib.spring.autoconfig AutoConfiguration by default to allow clients
+ * to opt-in to our own setup instead of needing to opt out when this library is added to their
+ * classpath.
+ *
+ * <p>We include the io.opencensus.contrib.spring package to make use of its @Traced annotation. It
```
gjuggler, line 11: Hah – so it's kind of a fight fire with fire approach here :) Thanks for leaving the detailed comment!

src/main/java/bio/terra/common/tracing/disable/AutoConfigDisable.java
```diff
@@ -0,0 +1,26 @@
...
+ * also, however, autoconfigures client and server trace annotations. We want to be able to make use
+ * of the @Traced annotation, but client services importing this library should not be
+ * autoconfigured. To do this, this sub package is EnableAutoConfiguratoin by
```
gjuggler: nit: spelling

src/main/java/bio/terra/common/tracing/TracingProperties.java
```diff
@@ -0,0 +1,44 @@
...
+public class TracingProperties {
+  /** Rate of trace sampling, 0.0 - 1.0 */
+  private double samplingRate = 0.1; // TODO DO NOT SUBMIT
+
+  /** Whether Stackdriver tracing export is enabled. */
+  private boolean stackdriverExportEnabled = true;
+
+  /**
+   * What HTTP URL patterns to enable tracing for. An empty list enables tracing for all HTTP
+   * requests.
```
gjuggler, line 21: Is there a reference we can point to for more details on the pattern syntax? It doesn't look like a regexp.

src/main/java/bio/terra/common/tracing/TracingProperties.java
```diff
@@ -0,0 +1,44 @@
+package bio.terra.common.tracing;
+
+import com.google.common.collect.Lists;
+import java.util.List;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "terra.common.tracing")
```
gjuggler, line 11: nit: add Javadoc for this public class. A simple line like "Defines config properties for the tracing package; see {@link TracingConfig} for their use." should be enough.

src/main/java/bio/terra/common/tracing/TracingProperties.java
```diff
@@ -0,0 +1,44 @@
...
+public class TracingProperties {
+  /** Rate of trace sampling, 0.0 - 1.0 */
+  private double samplingRate = 0.1; // TODO DO NOT SUBMIT
```
gjuggler: Is the TODO to decide on a default value here? I might expect 0.01, or 0.05 at most...

src/main/java/bio/terra/common/tracing/TracingProperties.java
```diff
@@ -0,0 +1,44 @@
...
+
+  /** Whether Stackdriver tracing export is enabled. */
+  private boolean stackdriverExportEnabled = true;
+
+  /**
+   * What HTTP URL patterns to enable tracing for. An empty list enables tracing for all HTTP
+   * requests.
+   */
+  private List<String> urlPatterns = Lists.newArrayList("/api/*");
```
gjuggler, line 25: Side note: I may end up adding a similar config to the logging package, since we have the same issue with very noisy but rarely useful /version and /status request logging. Alternatively, there it may make more sense to add a denylist of paths we want to exclude from logs unless it's a non-2xx response. Filed PF-547 to track that.

src/main/java/bio/terra/common/tracing/TracingAttributeAnnotatorInterceptor.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.web.method.HandlerMethod;
+import org.springframework.web.servlet.HandlerInterceptor;
+
+/**
+ * A {@link HandlerInterceptor} that adds more tracing attributes to HTTP requests.
+ *
+ * <p>This adds attributes that are Terra specific or could not be figured out by the {@link
+ * io.opencensus.contrib.http.servlet.OcHttpServletFilter}.
+ */
+class TracingAttributeAnnotatorInterceptor implements HandlerInterceptor {
```
gjuggler: [opt naming nit] this feels a tad verbose. One could easily drop the Tracing prefix since it's in the package name... and attribute / annotation are very similar in sound and meaning. Maybe `SpanAnnotationInterceptor`, or `RequestAttributeInterceptor`?

src/main/java/bio/terra/common/tracing/TracingAttributeAnnotatorInterceptor.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.slf4j.MDC;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.method.HandlerMethod;
+import org.springframework.web.servlet.HandlerInterceptor;
+
+/**
+ * A {@link HandlerInterceptor} that adds more tracing attributes to HTTP requests.
```
gjuggler: nit: it's worth being a bit more explicit that this interceptor is adding attributes to spans that we expect to have been already created by the OcHttpServletFilter. Here or in the next paragraph.

src/main/java/bio/terra/common/tracing/TracingAttributeAnnotatorInterceptor.java
```diff
@@ -0,0 +1,58 @@
...
+              Arrays.stream(mapping.path()).findFirst().orElse("unknown")));
+    }
+    // We expect the RequestIdFilter to have already set up the MDC context. This is true today
+    // because servlet Filters are always run before Spring HandlerInterceptors.
+    String requestId = MDC.get(RequestIdFilter.REQUEST_ID_MDC_KEY);
+    if (requestId != null) {
+      attributes.put("/terra/requestId", AttributeValue.stringAttributeValue(requestId));
+    }
+    attributes.put(
+        "/terra/operationId",
```
gjuggler: nit: I had to double-check why this is called `operationId`. Maybe worth a code comment mentioning that this traces back to the `operationId` field from OpenAPI 3 swagger definitions? (Mostly since that doesn't show up anywhere in the common library, so is not exactly locally self-documenting)

src/main/java/bio/terra/common/tracing/TracingConfig.java
```diff
@@ -0,0 +1,128 @@
...
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.web.servlet.FilterRegistrationBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+/**
+ * Auto-configuration class for Terra common tracing setup.
```
gjuggler: nit: technically this isn't auto-configuration. Best to avoid that term.

src/main/java/bio/terra/common/tracing/TracingConfig.java
```diff
@@ -0,0 +1,128 @@
...
+      attributes.put("/terra/component", AttributeValue.stringAttributeValue(componentName));
+    }
+    String version = configurableEnvironment.getProperty("spring.application.version");
+    if (version != null) {
+      attributes.put("/terra/version", AttributeValue.stringAttributeValue(version));
+    }
+    return attributes;
+  }
+
+  @Override
```
gjuggler: [opt method ordering nit] it may make sense to bump these two shorter methods up above the `afterPropertiesSet`-related methods, so they're less hidden below the fold.

src/main/java/bio/terra/common/tracing/TracingConfig.java
```diff
@@ -0,0 +1,128 @@
...
+  public FilterRegistrationBean<OcHttpServletFilter> tracingServletFilter() {
+    FilterRegistrationBean<OcHttpServletFilter> registration =
+        new FilterRegistrationBean(new OcHttpServletFilter());
+    registration.setUrlPatterns(tracingProperties.getUrlPatterns());
+    return registration;
+  }
+
+  @Override
+  public void afterPropertiesSet() {
+    setTraceSamplingProbability();
```
gjuggler, line 84: super nit: it felt a bit weird to see this broken out into a separate method but not the Stackdriver logic below, which is longer in length. Not sure it's worth changing, but wanted to note it.

src/test/java/bio/terra/common/tracing/TracedAnnotatedBean.java
```diff
@@ -0,0 +1,22 @@
...
+import io.opencensus.contrib.spring.aop.Traced;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.Tracing;
+import org.springframework.stereotype.Component;
+
+/** A bean with an {@link Traced} annotation for testing. */
+@Component
+class TracedAnnotatedBean {
+  /** The most recent span seen by {@link #annotatedMethod()}. */
+  private Span latestSpan;
```
gjuggler, line 12: Nice approach to storing the as-observed span info. I was wondering how you'd approach this part in the test.

src/test/java/bio/terra/common/tracing/TracingTest.java
```diff
@@ -0,0 +1,77 @@
...
+    RecordEventsSpanImpl beanSpan = (RecordEventsSpanImpl) annotatedBean.getLatestSpan();
+    SpanData beanSpanData = beanSpan.toSpanData();
+    assertEquals("annotatedMethod", beanSpanData.getName());
+    assertEquals(requestSpan.getContext().getSpanId(), beanSpanData.getParentSpanId());
+  }
+
+  /** Converts all types of {@link AttributeValue} to String. */
+  private static String coerceToString(AttributeValue attributeValue) {
+    String coerced =
+        attributeValue.match(
```
gjuggler, line 69: 🤦 just Java being Java here...

gjuggler (approved):
Looks really good! I like the approach you took to disabling the OpenCensus auto-config.

There is one thing missing here, which is an approach to creating spans for client API calls and/or downstream propagation of trace context. It may make most sense to file that as a follow-up ticket, given this is in good shape as-is.

I don't think there's a great way to automatically set up tracing & propagation for arbitrary API clients, but we can probably whittle it down to a single utility method that gets called at the service level with each API client instance. I'd tinkered a bit with this in a WSM branch (see [code](https://github.com/DataBiosphere/terra-workspace-manager/blob/1ea57e5e840f17f330bc95b4146a7e7b1af65705/src/main/java/bio/terra/workspace/service/buffer/BufferService.java)).

wchamber:
> Looks really good! I like the approach you took to disabling the OpenCensus auto-config.
> 
> There is one thing missing here, which is an approach to creating spans for client API calls and/or downstream propagation of trace context. It may make most sense to file that as a follow-up ticket, given this is in good shape as-is.
> 
> I don't think there's a great way to automatically set up tracing & propagation for arbitrary API clients, but we can probably whittle it down to a single utility method that gets called at the service level with each API client instance. I'd tinkered a bit with this in a WSM branch (see [code](https://github.com/DataBiosphere/terra-workspace-manager/blob/1ea57e5e840f17f330bc95b4146a7e7b1af65705/src/main/java/bio/terra/workspace/service/buffer/BufferService.java)).

Much prefer to do as follow up work. Filed PF-548. I wonder if we used a different openapi client code generator from jersey we might be able to plug in better with Spring beans and filters and do central registration.

src/main/java/bio/terra/common/tracing/TracingProperties.java
wchamber: Yep, thanks. Went with 0.05.

src/main/java/bio/terra/common/tracing/TracingProperties.java
wchamber: SG. FWIW, I didn't see any denylist on FilterRegistrationBean, just an allowlist.

src/test/java/bio/terra/common/tracing/TracingTest.java
wchamber: extra Java fun for your enjoyment: this `AttribueValue#match` function wants 5 functions for each of String, Boolean, Long, Double, Object. The class specific `toString` overload with method reference operator works for `String::toString`, but not for Boolean, Long or Double due to overloads on those classes of `toString`. So therefore 5 `Object::toString` method references :)

src/main/java/bio/terra/common/tracing/TracingConfig.java
wchamber: done, updated LoggingConfig as well :)

src/main/java/bio/terra/common/tracing/TracingAttributeAnnotatorInterceptor.java
wchamber: done

src/main/java/bio/terra/common/tracing/TracingAttributeAnnotatorInterceptor.java
wchamber: done, ptal

src/main/java/bio/terra/common/tracing/TracingAttributeAnnotatorInterceptor.java
wchamber: Thanks, like RequestAttributeInterceptor.
Traces also have annotations, which are timestamped events distinct from attributes, so not going with anything `annotation`.

src/test/java/bio/terra/common/tracing/TracedAnnotatedBean.java
wchamber: The RecordEventsSpanImpl cast in the tests is a little gross, but seemed preferable to trying to set up a custom exporter and waiting for it to export spans.

src/main/java/bio/terra/common/tracing/TracingProperties.java
wchamber: Added a comment link to https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/pattern/PathPattern.html

src/main/java/bio/terra/common/tracing/RequestAttributeInterceptor.java
```diff
@@ -0,0 +1,62 @@
...
+ * io.opencensus.contrib.http.servlet.OcHttpServletFilter}.
+ */
+class RequestAttributeInterceptor implements HandlerInterceptor {
+  @Override
+  public boolean preHandle(
+      HttpServletRequest httpRequest, HttpServletResponse httpResponse, Object handler) {
+    if (handler instanceof HandlerMethod) {
+      HandlerMethod handlerMethod = (HandlerMethod) handler;
+      // This relies on the span for the request being already set as the current span. If there is
+      // no span, this is a no-op.
```
jmczerk, line 33: Do we expect a span to be present in the general case?  Just wondering how often we'll build an attributes map that is not consumed.

src/main/java/bio/terra/common/tracing/TracingProperties.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Defines configuration properties for the tracing package; see {@link TracingConfig} for their
+ * use.
+ */
+@ConfigurationProperties(prefix = "terra.common.tracing")
+public class TracingProperties {
+  /** Rate of trace sampling, 0.0 - 1.0 */
+  private double samplingRate = 0.05; // TODO DO NOT SUBMIT
```
jmczerk: `TODO DO NOT SUBMIT`?

jmczerk (approved):
Very clean and great docs, had a couple questions but otherwise LGTM.

src/main/java/bio/terra/common/tracing/RequestAttributeInterceptor.java
wchamber: We expect there to be a span for "interesting" requests, though in WM today we get more request to the /status endpoint than anything else. We don't expect the Span to be sampled very often. So, we probably often build a map that doesn't have any meaningful effect. But the Span API doesn't really try to expose any way to determine if the Span is going to be sampled, so trying to figure it out is discouraged.
If this ever shows up on a performance profile, we could revisit.

src/main/java/bio/terra/common/tracing/TracingProperties.java
wchamber: argh, thanks. Was tweaking the value, but forgot to remove the comment.

PR #29, opened 2021-03-11
PF-279 Don't fail on re-registration of StackdriverTraceExporter.

wchamber:
In client tests, the StackdriverTraceExporter can be registered multiple times by multiple instances TracingConfig bean. This is ok in testing.
This is hard to reproduce in terra-common-lib because we don't yet have any google credentials to allow registration to succeed, so the StackdriverTraceExporter does not get registered successfully at all, much less multiple times. I don't think it's worth adding google credentials for testing for this use case, though we may eventually want them.

Found trying to integrate with Workspace Manager.


jmczerk (approved):
LGTM

gjuggler (approved):

PR #30, opened 2021-03-12
[PF-539] Add support for logging arbitrary Gson-type JSON nodes in structured payload.

gjuggler:
This is one of the precursor steps to rolling out the TCL logging package to RBS.

The overall goal is to make it so RBS produces structured logs for CRL-initiated actions, including full request payload and any Google exception data.

We should be able to use the "structured payload" approach that the TCL logging package implements, where a service can simply write

```
LOG.info("message", structuredData)
```

and the structuredData will show up as part of the `jsonPayload` in Cloud Logging.

However, one wrinkle in the implementation is that the logging library uses Jackson for JSON serialization, while CRL uses Gson. The path of least resistance here seemed to be to explicitly support a `gson.JsonObject` in the logging payload, similar to how we already support a Jackson-style `jackson.databind.JsonNode`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ repositories {
 }
 
 group 'bio.terra'
-version '0.0.22-SNAPSHOT'
+version '0.0.25-SNAPSHOT'
 sourceCompatibility = JavaVersion.VERSION_1_8
 
 def OPENCENSUS_VERSION = '0.28.+'
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ repositories {
 }
 
 group 'bio.terra'
-version '0.0.25-SNAPSHOT'
+version '0.0.22-SNAPSHOT'
 sourceCompatibility = JavaVersion.VERSION_1_8
 
 def OPENCENSUS_VERSION = '0.28.+'
--- a/src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
+++ b/src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
@@ -133,7 +133,7 @@ class GoogleJsonLayout extends JsonLayoutBase<ILoggingEvent> {
     // All MDC properties will be directly splatted onto the JSON map. This is how the MDC
     // 'requestId' property ends up in the JSON output, and ultimately into jsonPayload.requestId
     // in cloud logging.
-    event.getMDCPropertyMap().forEach(outputMap::put);
+    outputMap.putAll(event.getMDCPropertyMap());
 
     // Generically splat any map-like or JSON-like argument to the log call onto the output JSON.
     // This is how e.g. the RequestLoggingFilter adds the 'httpRequest' object to the JSON output.
@@ -169,7 +169,7 @@ class GoogleJsonLayout extends JsonLayoutBase<ILoggingEvent> {
 
     // If the generic JSON splatting above caused a 'labels' entry to exist, move the value to
     // the well-known key that Cloud Logging will ingest as proper labels key-value pairs.
-    if (outputMap.get("labels") != null) {
+    if (outputMap.containsKey("labels")) {
       outputMap.put("logging.googleapis.com/labels", outputMap.get("labels"));
       outputMap.remove("labels");
     }
--- a/src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
+++ b/src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
@@ -107,33 +107,33 @@ class GoogleJsonLayout extends JsonLayoutBase<ILoggingEvent> {
    */
   @Override
   protected Map<String, Object> toJsonMap(ILoggingEvent event) {
-    Map<String, Object> map = new LinkedHashMap<>();
+    Map<String, Object> outputMap = new LinkedHashMap<>();
 
-    map.put("timestampSeconds", TimeUnit.MILLISECONDS.toSeconds(event.getTimeStamp()));
-    map.put("timestampNanos", TimeUnit.MILLISECONDS.toNanos(event.getTimeStamp() % 1_000));
+    outputMap.put("timestampSeconds", TimeUnit.MILLISECONDS.toSeconds(event.getTimeStamp()));
+    outputMap.put("timestampNanos", TimeUnit.MILLISECONDS.toNanos(event.getTimeStamp() % 1_000));
 
-    map.put("severity", String.valueOf(event.getLevel()));
-    map.put("message", getMessage(event));
+    outputMap.put("severity", String.valueOf(event.getLevel()));
+    outputMap.put("message", getMessage(event));
     Map<String, Object> serviceContextMap = new HashMap<>();
     serviceContextMap.put(
         "service", applicationContext.getEnvironment().getProperty("spring.application.name"));
     serviceContextMap.put(
         "version", applicationContext.getEnvironment().getProperty("spring.application.version"));
-    map.put("serviceContext", serviceContextMap);
+    outputMap.put("serviceContext", serviceContextMap);
 
-    map.put("context", event.getLoggerContextVO().getName());
-    map.put("thread", event.getThreadName());
-    map.put("logger", event.getLoggerName());
-    map.put("logging.googleapis.com/sourceLocation", getSourceLocation(event));
+    outputMap.put("context", event.getLoggerContextVO().getName());
+    outputMap.put("thread", event.getThreadName());
+    outputMap.put("logger", event.getLoggerName());
+    outputMap.put("logging.googleapis.com/sourceLocation", getSourceLocation(event));
 
-    addTraceId(map);
-    addSpanId(map);
-    addTraceSampled(map);
+    addTraceId(outputMap);
+    addSpanId(outputMap);
+    addTraceSampled(outputMap);
 
     // All MDC properties will be directly splatted onto the JSON map. This is how the MDC
     // 'requestId' property ends up in the JSON output, and ultimately into jsonPayload.requestId
     // in cloud logging.
-    event.getMDCPropertyMap().forEach(map::put);
+    event.getMDCPropertyMap().forEach(outputMap::put);
 
     // Generically splat any map-like or JSON-like argument to the log call onto the output JSON.
     // This is how e.g. the RequestLoggingFilter adds the 'httpRequest' object to the JSON output.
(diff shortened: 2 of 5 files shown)


src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -133,11 +141,21 @@ public void stop() {
       for (Object arg : event.getArgumentArray()) {
         try {
           if (arg instanceof Map) {
+            // Handle arbitrary Map by splatting each key-value pair into the main output map.
```
jaycarlton, line 144: nit: you can use map.putAll(jsonMap) below.

yonghaoy (approved):
Thanks! The reason of choosing Gson was not widely discussed. By the time we implement CRL, I just did a google search "Jackson vs Gson", and seems Gson seems to be a better option for me. (easy to use, no checked exception everywhere). But I didn't consider that Spring is tightly with Jackson. I am fine to switch CRL to use Jackson if you think that is McTerra tech stack.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
jaycarlton: rename `map` to `outputMap` at line 110 for clarity.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -133,11 +141,21 @@ public void stop() {
...
             jsonNode.fields().forEachRemaining(entry -> map.put(entry.getKey(), entry.getValue()));
+          } else if (arg instanceof JsonObject) {
+            // Some libraries use GSON rather than Jackson for arbitrary JSON data, and we should
+            // support that too. Since we're using Jackson for top-level serialization, we need to
+            // convert this gson.JsonObject into a jackson.databind.JsonNode for storage in the
+            // output map. The simplest way to do this is by using a JSON string as intermediary.
+            JsonObject jsonObject = (JsonObject) arg;
+            JsonNode jsonNode = objectMapper.readTree(jsonObject.toString());
+            jsonNode.fields().forEachRemaining(entry -> map.put(entry.getKey(), entry.getValue()));
           }
```
jaycarlton, line 163: Log/warn if an unexpected type shows up here. Indicate if there are other types that intentionally aren't handled.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -57,18 +60,23 @@
  */
 class GoogleJsonLayout extends JsonLayoutBase<ILoggingEvent> {
 
-  private Tracer tracer = Tracing.getTracer();
-
   // A reference to the current Spring app context, on order to pull out the spring.application.name
   // and spring.application.version variable for inclusion in JSON output.
   private ConfigurableApplicationContext applicationContext;
```
jaycarlton, line 65: It's nice that you're able to get JSON into stackdriver without using using their logging client everywhere.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -133,11 +141,21 @@ public void stop() {
       for (Object arg : event.getArgumentArray()) {
         try {
           if (arg instanceof Map) {
+            // Handle arbitrary Map by splatting each key-value pair into the main output map.
             Map<String, Object> jsonMap = (Map<String, Object>) arg;
             jsonMap.forEach(map::put);
           } else if (arg instanceof JsonNode) {
+            // Handle Jackson JsonNode by splatting each property sub-tree into the main output map.
             JsonNode jsonNode = (JsonNode) arg;
             jsonNode.fields().forEachRemaining(entry -> map.put(entry.getKey(), entry.getValue()));
```
jaycarlton: Do object/composite/nested properties get expanded into JSON as well here? Seems like you'd have to recurse through more than on level. Looks like the value here is another JsonNode according to the docs.

src/test/java/bio/terra/common/logging/LoggingTestController.java
```diff
@@ -40,5 +41,14 @@ public void testStructuredLogging() throws JsonProcessingException {
...
     LOG.info("Structured data", LoggingUtils.structuredLogData("pojo", pojo));
+
+    // Test that a raw GSON JsonObject works too. Some libraries such as CRL may prefer to include
+    // GSON type objects in the log event payload.
+    JsonObject jsonObject = new JsonObject();
+    JsonObject innerObject = new JsonObject();
+    innerObject.addProperty("bar", "baz");
+    jsonObject.add("foo", innerObject);
+    // The GSON should look like {"foo": {"bar": "baz"}}
+    LOG.info("GSON object", jsonObject);
```
jaycarlton, line 56: can you write an assertion for this?

build.gradle
```diff
@@ -22,7 +22,7 @@ repositories {
 }
 
 group 'bio.terra'
-version '0.0.22-SNAPSHOT'
+version '0.0.25-SNAPSHOT'
```
yonghaoy: nit: this is optional. GHA can figure this out without this change. (it reads https://github.com/DataBiosphere/terra-common-lib/tags and bump based on that)

yonghaoy (approved):

build.gradle
gjuggler: Ack – it was just helpful in my local testing, to make sure my local Maven publishing used the right version.

gjuggler:
> Thanks! The reason of choosing Gson was not widely discussed. By the time we implement CRL, I just did a google search "Jackson vs Gson", and seems Gson seems to be a better option for me. (easy to use, no checked exception everywhere). But I didn't consider that Spring is tightly with Jackson. I am fine to switch CRL to use Jackson if you think that is McTerra tech stack.

Thanks for the context! I'm not too worried about backfilling consistency in CRL, but I suspect we'll want a general trend of defaulting to Jackson for JSON manipulation unless we have a good reason not to.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
gjuggler: Done and done.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
gjuggler: Yes – the `JsonLayoutBase` class handles using the `JsonFormatter` (which we register in the constructor as a `JacksonJsonFormatter`) to format each map entry as a JSON element. This is exercised in the unit tests.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
gjuggler: This isn't meant to handle all argument types. Most arguments to a logback call are strings or Exceptions, which we want to ignore (since other parts of the logback machinery will handle those).

I'd meant to convey this approach in the comment at the top of the if statement ("Generically splat any map-like or JSON-like argument..."). Is there a better place to describe it?

src/test/java/bio/terra/common/logging/LoggingTestController.java
gjuggler: Yes – see the tests themselves in LoggingTest. I'll add a class-level comment making clear what role this class plays.

build.gradle
jaycarlton: Is it possible to use a local version string to make sure it doesn't get checked in? Like `0.0.1-LOCAL`?

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -99,45 +107,59 @@ public void stop() {
...
+    outputMap.put("timestampSeconds", TimeUnit.MILLISECONDS.toSeconds(event.getTimeStamp()));
+    outputMap.put("timestampNanos", TimeUnit.MILLISECONDS.toNanos(event.getTimeStamp() % 1_000));
 
-    map.put("severity", String.valueOf(event.getLevel()));
-    map.put("message", getMessage(event));
+    outputMap.put("severity", String.valueOf(event.getLevel()));
+    outputMap.put("message", getMessage(event));
     Map<String, Object> serviceContextMap = new HashMap<>();
     serviceContextMap.put(
         "service", applicationContext.getEnvironment().getProperty("spring.application.name"));
```
jaycarlton, line 119: Can this class not be autowired? You could use `Provider<String>` for these properties I think.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -99,45 +107,59 @@ public void stop() {
...
-    addTraceSampled(map);
+    addTraceId(outputMap);
+    addSpanId(outputMap);
+    addTraceSampled(outputMap);
 
     // All MDC properties will be directly splatted onto the JSON map. This is how the MDC
     // 'requestId' property ends up in the JSON output, and ultimately into jsonPayload.requestId
     // in cloud logging.
-    event.getMDCPropertyMap().forEach(map::put);
+    event.getMDCPropertyMap().forEach(outputMap::put);
```
jaycarlton: nit: putAll()

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
jaycarlton: For my own confusion, maybe just an example of what happens to a JsonNode. Does it become a `Map<String, JsonNode>`? If so, how is that any better?

And you're not trying to convert all POJO types into JSON, which I originally thought was a goal for some reason.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -147,12 +169,12 @@ public void stop() {
 
     // If the generic JSON splatting above caused a 'labels' entry to exist, move the value to
     // the well-known key that Cloud Logging will ingest as proper labels key-value pairs.
-    if (map.get("labels") != null) {
-      map.put("logging.googleapis.com/labels", map.get("labels"));
-      map.remove("labels");
+    if (outputMap.get("labels") != null) {
```
jaycarlton: use `containsKey("labels")`

jaycarlton (approved):
Thanks for doing this.

build.gradle
jmczerk: In a past project we used something like this (not sure it was this exact plugin), such that only major/minor changes were changed in source control, and patch bumps happened implicitly: https://plugins.gradle.org/plugin/eu.davidea.grabver

gjuggler: Thanks John – I added this option to the [turpentine log](https://docs.google.com/document/d/1JTQnS0FmuboW31wmmGMcjWww7wXix8ODbqcvkbQbtp4/edit#) doc.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
gjuggler: The logging stuff gets constructed before the application context is fully initialized, so there are some odd quirks around what amount of bean configuration is available by then. I wouldn't dig any deeper here, given it's pretty simple constructor dependency injection and it works.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
gjuggler: Done

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
gjuggler: The best way I found to describe it is in the comment above the prior if-else: "splat each property sub-tree into the main output map".

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
gjuggler: Done

PR #31, opened 2021-03-12
[PF-456] StairwayComponent and its dependencies

jaycarlton:
Move the `StairwayComponent` from buffer into the common lib.

Corresponding PR in buffer: https://github.com/DataBiosphere/terra-resource-buffer/pull/128


src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -0,0 +1,138 @@
...
+    Stairway.Builder builder =
+        Stairway.newBuilder()
+            .maxParallelFlights(stairwayConfiguration.getMaxParallelFlights())
+            .applicationContext(applicationContext)
+            .keepFlightLog(true)
+            .stairwayName(kubeService.getPodName())
+            .stairwayClusterName(stairwayClusterName)
+            .workQueueProjectId(getDefaultProjectId())
+            .enableWorkQueue(kubernetesComponent.isInKubernetes())
+            .stairwayHook(new TracingHook());
```
yonghaoy: At least the hook should be customizable, not sure if we need to customize other configs.

src/main/java/bio/terra/common/stairway/StairwayJdbcConfiguration.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.common.stairway;
+
+import bio.terra.common.db.JdbcConfiguration;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "buffer.stairway.db")
```
yonghaoy: prefix should be the `common.stairway.db`

yonghaoy (changes requested):

src/main/java/bio/terra/common/stairway/StairwayJdbcConfiguration.java
wchamber: `terra.common.stairway.db` if we want to match the tracing package.

src/main/java/bio/terra/common/db/JdbcConfiguration.java
```diff
@@ -0,0 +1,82 @@
...
+      configureDataSource();
+    }
+    return dataSource;
+  }
+
+  private void configureDataSource() {
+    Properties props = new Properties();
+    props.setProperty("user", getUsername());
+    props.setProperty("password", getPassword());
+
```
jmczerk: You might consider adding the ability to configure `apache.commons.pool2` JMX integration to avoid issues like PF-485:

https://github.com/DataBiosphere/terra-resource-janitor/pull/93

ddietterich: I don't think we want to use dbcp2 for our pooling.
We don't need to do it in this work, but we should change over to HikariCP.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -0,0 +1,138 @@
...
+        "Creating Stairway: name: [{}]  cluster name: [{}]",
+        kubeService.getPodName(),
+        stairwayClusterName);
+
+    // TODO(PF-314): Cleanup old flightlogs.
+    // TODO(PF-456):
+    Stairway.Builder builder =
+        Stairway.newBuilder()
+            .maxParallelFlights(stairwayConfiguration.getMaxParallelFlights())
+            .applicationContext(applicationContext)
```
ddietterich: The Stairway applicationContext is type Object. That should be overridable. For example, in WSM, we use the FlightBeanBag for the context, not the Spring ApplicationContext.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -0,0 +1,138 @@
...
+  @Autowired
+  public StairwayComponent(
+      ApplicationContext applicationContext,
+      StairwayConfiguration stairwayConfiguration,
+      StairwayJdbcConfiguration stairwayJdbcConfiguration,
+      KubernetesComponent kubernetesComponent) {
+    this.stairwayConfiguration = stairwayConfiguration;
+    this.stairwayJdbcConfiguration = stairwayJdbcConfiguration;
+    this.kubeService = kubernetesComponent.get();
+    String stairwayClusterName = kubeService.getNamespace() + "buffer--stairwaycluster";
```
ddietterich: I think cluster name should be parameterized and not derived from K8s.
If you do want to default it, then probably don't  call it "buffer"

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -0,0 +1,138 @@
...
+            .keepFlightLog(true)
+            .stairwayName(kubeService.getPodName())
+            .stairwayClusterName(stairwayClusterName)
+            .workQueueProjectId(getDefaultProjectId())
+            .enableWorkQueue(kubernetesComponent.isInKubernetes())
+            .stairwayHook(new TracingHook());
+    try {
+      stairway = builder.build();
+    } catch (StairwayExecutionException e) {
+      throw new IllegalArgumentException("Failed to build Stairway.", e);
```
ddietterich: What is the purpose of the exception conversion?

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -0,0 +1,138 @@
...
+@Component
+public class StairwayComponent {
+  private final Logger logger = LoggerFactory.getLogger(StairwayComponent.class);
+
+  private final StairwayConfiguration stairwayConfiguration;
+  private final StairwayJdbcConfiguration stairwayJdbcConfiguration;
+  private final Stairway stairway;
+  private final KubeService kubeService;
+
+  public enum Status {
```
ddietterich, line 32: What purpose is served by this Status tracking?
I would prefer it were deleted.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
jaycarlton: How do I make that customizable? As a constructor parameter, like a `Provider<TracingHook>`?

src/main/java/bio/terra/common/db/JdbcConfiguration.java
jaycarlton: I'm not familiar with JMX. I may wait on this.

src/main/java/bio/terra/common/stairway/StairwayJdbcConfiguration.java
jaycarlton: `terra.common.stairway.db`?

src/main/java/bio/terra/common/stairway/StairwayComponent.java
ddietterich: I think the only way to make this and applicationContext customizable is to allow them to be overridden on the `initialize()` call in Stairway. Otherwise, you'd need to be able to autowire some bean provided by the component that would provide the values for the tracing and context. It gets pretty tangled.

src/main/java/bio/terra/common/db/JdbcConfiguration.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.commons.dbcp2.DriverManagerConnectionFactory;
+import org.apache.commons.dbcp2.PoolableConnection;
+import org.apache.commons.dbcp2.PoolableConnectionFactory;
+import org.apache.commons.dbcp2.PoolingDataSource;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+import org.apache.commons.pool2.ObjectPool;
+import org.apache.commons.pool2.impl.GenericObjectPool;
+
+public class JdbcConfiguration {
```
jaycarlton: I'm realizing that pulling in this class expands the scope considerably, if I'm to make it appropriate for every service (independently of the StairwayComponent). I'm not sure if I should tackle that or not.

src/main/java/bio/terra/common/kubernetes/KubernetesComponent.java
```diff
@@ -0,0 +1,30 @@
+package bio.terra.common.kubernetes;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/** A Spring Component for exposing an initialized {@link KubeService}. */
+@Component
+public class KubernetesComponent {
```
gjuggler: [General comment on Spring config & properties management, with a suggestion on a potential simplification here.]

We're still working out our exact conventions, but I thought Will's approach to Spring config management in the tracing package ([link](https://www.baeldung.com/spring-enable-config-properties)) was pretty refined. In short, it contains:

- A single @FooConfig class, whose job is to expose configured beans and apply additional configurations / customizations to the Spring application environment. FooConfig has an @EnableConfigurationProperties annotation which points to...
- A @FooProperties class, which defines a simple POJO representing structured properties supporting externalized configuration for this package.

See https://www.baeldung.com/spring-enable-config-properties for a summary of why this approach can be nice.

With that model, I don't think a KubeComponent class would be strictly necessary. Consider KubeConfig as the primary class for configuring and providing beans to the environment. It would have an Autowired KubeProperties instance and could expose KubeService as a @Bean annotated method:

```
@Bean
public KubeService getKubeService() {
  return new KubeService(kubeProperties.getPodName(), ...);
}
```

and then from `StairwayComponent` you can Autowire a KubeService rather than KubeComponent.

With this framing, `StairwayConfiguration` would become `StairwayProperties`, and you could probably include it in the @EnableConfigurationProperties annotation in the main config class.

gjuggler (commented):
One comment from the peanut gallery on Spring config conventions. I'm somewhat channeling Will here, as he and I both went through some {soul-searching, head-scratching} in the logging and tracing packages. Happy to chat more offline if my comment isn't super clear.

src/main/java/bio/terra/common/kubernetes/KubernetesComponent.java
gjuggler: And one addendum: I understand the strong motivation to keep `StairwayComponent` API-compatible with the existing `StairwayComponent` built into each service repo, to ease the transition. Makes sense. But maybe a simple rename would make its purpose more clear, i.e. `StairwayLifecycleManager` or `StairwayInitializer` or `StairwayKubeAdapter`? In the context of this general package, the -Component suffix doesn't add much.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
yonghaoy: We may want to pass `List<StairwayHook>`. 
One possible solution is the constructor just build Stairway.Builder. Then change the following method to `initialize(List<StairwayHook> stairwayHooks)`

yonghaoy: For example: WSM needs two hooks:
https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/service/job/JobService.java#L86

src/main/java/bio/terra/common/kubernetes/KubernetesComponent.java
jaycarlton: I like it.  I thought "-Component" was becoming the pattern for this service-provider-plus-properties class. Step 1 was just jamming the class in here and seeing if I could make it work. It sounded like I shouldn't inject a `Provider<Stairway>` (and couldn't inject Stairway directly b/c it's not a bean). I definitely favor keeping properties and config and the service wrapper/provider separate. @yonghaoy what do you think?

src/main/java/bio/terra/common/stairway/StairwayJdbcConfiguration.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.common.stairway;
+
+import bio.terra.common.db.JdbcConfiguration;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
```
jaycarlton: I want to consolidate this into the parent class and rename that, as I'm not interested in it being the same code as in buffer service. That is, I don't want to be in the business of prescribing JDBC connecting for other services yet.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -0,0 +1,138 @@
...
+      logger.info("Request Stairway terminate");
+      shutdownSuccess =
+          stairway.terminate(
+              stairwayConfiguration.getTerminateTimeout().toMillis(), TimeUnit.MILLISECONDS);
+    }
+    logger.info("Finished Stairway shutdown?: {}", shutdownSuccess);
+    return shutdownSuccess;
+  }
+
+  public Stairway get() {
```
jaycarlton, line 133: @gjuggler so this becomes a `Provider<Stairway>` that's separately provided by the new config class and autowired?

src/main/java/bio/terra/common/stairway/StairwayComponent.java
jaycarlton: Why not `Provider<List<StairwayHook>>` in the constructor? If I build the builder in the ctor, how do I add the hooks to it in `initialize()`?

Edit: I guess any properties that are service-specified are going to need to come in via the yaml files; I can't define another properties class or arrange for arbitrary beans to be provided with this approach. Or can I ask for a bean with a given name in code in the common lib but not have that resolved unless a service provides a bean with that name?

jaycarlton (commented):
Good inputs. I have some questions.

src/main/java/bio/terra/common/db/JdbcConfiguration.java
yonghaoy: Pulling this class seems will make the migration easier, and I think it makes sense to pull this class here.
For migrating client to using this base class, it does sound out of this ticket's scope. You can create few tickets to track clients migration for using this base class, and for this ticket, just use if for `StairwayJdbcConfiguration` migration.  `ServerJdbcConfiguration` migration should be straight forward, anyone can pick it up.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
yonghaoy: that SG!

src/main/java/bio/terra/common/db/JdbcConfiguration.java
jmczerk: One thing to note is that my comment re: JMX/PF-485 comes back into play if we go this route; if we consumed this as-is in Janitor it could re-introduce that defect there (for instance); so the follow-on ticket may then involve coming back to the common lib to add the configurable.

jaycarlton: Is the configuration we'd need something we can just add to the right properties file, or more interesting?

jmczerk: The scope of the code change can be seen in this commit (add a property and use it to setup the `GenericObjectPoolConfig<>` passed to `GenericObjectPool <>` ctor):

https://github.com/DataBiosphere/terra-resource-janitor/pull/93/commits/331e57c8dcc7dd07471fab7d7a916f3b61ba64a4

Downstream consumers would be able to consume it as needed (note that the extending class is the one that sets the config, thus the changes to two different test configs in that commit).

build.gradle
```diff
@@ -47,15 +49,15 @@ dependencies {
     implementation group: 'org.springframework' , name: 'spring-core'
     implementation group: 'org.springframework.boot', name: 'spring-boot-starter-data-jdbc'
     implementation group: 'org.springframework.boot', name: 'spring-boot-starter-web'
-
+    // https://mvnrepository.com/artifact/javax.inject/javax.inject
+    implementation group: 'javax.inject', name: 'javax.inject', version: '1'
```
jaycarlton: This is for Provider, which I may not need.

build.gradle
```diff
@@ -181,3 +183,7 @@ artifactory {
         }
     }
 }
+
+java {
+    withSourcesJar() // for publish plugin
```
jaycarlton: trying to get the publish plugin to include source code. I think it helps

src/main/java/bio/terra/common/db/JdbcConfig.java
```diff
@@ -0,0 +1,58 @@
...
+  private final JdbcProperties jdbcProperties;
+  // Not a property
+  private PoolingDataSource<PoolableConnection> dataSource;
+
+  public JdbcConfig(JdbcProperties jdbcProperties) {
+    this.jdbcProperties = jdbcProperties;
+  }
+
+  // Main use of the configuration is this pooling data source object.
+  @Bean
```
jaycarlton: I'm not sure if this is a good idea or not. I'm not sure how to have multiple data sources (stairway and buffer) if I go this route

src/main/java/bio/terra/common/db/JdbcProperties.java
```diff
@@ -0,0 +1,46 @@
+package bio.terra.common.db;
+
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "terra.common.db")
+public class JdbcProperties {
```
jaycarlton: This class used to be the `JdbcConfiguration` IIRC

src/main/java/bio/terra/common/db/JdbcProperties.java
```diff
@@ -0,0 +1,46 @@
...
+
+  public void setUsername(String username) {
+    this.username = username;
+  }
+
+  public void setPassword(String password) {
+    this.password = password;
+  }
+
+  @Override
```
jaycarlton: This isn't really a property, but it's not hurting anything at the moment.

src/main/java/bio/terra/common/kubernetes/KubeConfig.java
```diff
@@ -0,0 +1,21 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
+
+@Configuration
+@EnableConfigurationProperties(KubeProperties.class)
+public class KubeConfig {
+
+  @Bean
+  @Scope(value = ConfigurableBeanFactory.SCOPE_SINGLETON)
+  public KubeService getKubeService(KubeProperties kubernetesProperties) {
```
jaycarlton: Is it safe to do this without a `Provider`? I guess this function replaces the missing autowired constructor from KubeService.

src/main/java/bio/terra/common/stairway/StairwayConfig.java
```diff
@@ -0,0 +1,59 @@
...
+  @Qualifier("stairway.cluster.name")
+  public String getStairwayClusterName(
+      KubeService kubeService, KubeProperties kubernetesProperties) {
+    return kubeService.getNamespace()
+        + Optional.ofNullable(kubernetesProperties.getNamespaceSuffix()).orElse(StringUtils.EMPTY)
+        + "--stairwaycluster";
+  }
+
+  @Bean
+  public Stairway getStairway(
```
jaycarlton: This function is giving me trouble. The autowiring works but  then Stairway throws because of missing JSON or something. Still tracking this down.

src/main/java/bio/terra/common/stairway/StairwayJdbcProperties.java
```diff
@@ -0,0 +1,7 @@
+package bio.terra.common.stairway;
+
+import bio.terra.common.db.JdbcProperties;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "terra.common.stairway.db")
```
jaycarlton: Here I'm attempting to define one properties class identical to another but with a different prefix. I guess it _could_ work. We need to do the same thing with `BufferJdbcProperties`.

src/main/java/bio/terra/common/stairway/StairwayLifecycleManager.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  private final AtomicReference<Status> status = new AtomicReference<>(Status.INITIALIZING);
+
+  @Autowired
+  public StairwayLifecycleManager(
+      @Qualifier("stairway.cluster.name") String clusterName,
+      KubeService kubeService,
+      PoolingDataSource<PoolableConnection> dataSource,
+      Stairway stairway,
```
jaycarlton: THis class no longer builds Stairway itself but lets the configuration do that.

jaycarlton (commented):
Here's a snapshot of progress. Still broken in a couple of ways.

src/main/java/bio/terra/common/db/JdbcConfig.java
```diff
@@ -0,0 +1,58 @@
+package bio.terra.common.db;
```
ddietterich: I suggest putting Stairway's jdbc properties into the stairway properties and not trying to share the JdbcConfiguration here. It is only three things. If we want to share code for building the connection pool, that can be library methods.

src/main/java/bio/terra/common/kubernetes/KubeConfig.java
ddietterich: Does the kubernetesProperties bean get resolved right?

src/main/java/bio/terra/common/stairway/StairwayConfig.java
```diff
@@ -0,0 +1,59 @@
+package bio.terra.common.stairway;
```
ddietterich, line 1: What is the naming convention here? I understand the properties. The `Config` is not making sense to me.

src/main/java/bio/terra/common/db/JdbcProperties.java
```diff
@@ -0,0 +1,46 @@
+package bio.terra.common.db;
+
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "terra.common.db")
```
yonghaoy: JdbcProperties is supposed to be the common class that would be used directly by each service, as well as a super class of `StairwayJdbcProperties`
StairwayJdbcProperties overried the prefix is okay, but not sure if that is the best way of doing that. Or we want a "common super class" on top of that, and is not a `ConfigurationProperties`

src/main/java/bio/terra/common/stairway/StairwayJdbcProperties.java
ddietterich: We do something similar in TDR. I think it works. What is not clear to me is what other JdbcProperties will be in the common lib. The other databases are per-service and won't live here.

src/main/java/bio/terra/common/stairway/StairwayConfig.java
```diff
@@ -0,0 +1,59 @@
...
+  private final StairwayProperties stairwayProperties;
+  private final KubeProperties kubeProperties;
+
+  public StairwayConfig(StairwayProperties stairwayProperties, KubeProperties kubeProperties) {
+    this.stairwayProperties = stairwayProperties;
+    this.kubeProperties = kubeProperties;
+  }
+
+  @Bean
+  @Qualifier("stairway.cluster.name")
```
ddietterich: What is that used for?

src/main/java/bio/terra/common/stairway/StairwayConfig.java
```diff
@@ -0,0 +1,59 @@
...
+      KubeService kubeService, KubeProperties kubernetesProperties) {
+    return kubeService.getNamespace()
+        + Optional.ofNullable(kubernetesProperties.getNamespaceSuffix()).orElse(StringUtils.EMPTY)
+        + "--stairwaycluster";
+  }
+
+  @Bean
+  public Stairway getStairway(
+      ApplicationContext applicationContext,
+      @Qualifier("stairway.cluster.name") Provider<String> stairwayClusterNameProvider) {
```
ddietterich: I think if we have to use Provider we are doing it wrong. I would rather move setting in Stairway from construction to initialization.

In fact, I would prefer a setup without a StairwayConfig and just with the lifecycle manager. Let it construct Stairway given the properties and initialize Stairway (having moved the settings to initialization).

The (eventually common) JobService can autowire StairwayLifecycleManager and use a getter for performing flight submits and status checks.

src/main/java/bio/terra/common/db/JdbcProperties.java
jaycarlton: Yes, maybe I'll rename this `BaseJdbcProperties` and put a comment that clients should extend the class and put `@ConfigurationProperties` with the appropriate prefix.

ddietterich (commented):
I think this can be simplified.

src/main/java/bio/terra/common/db/JdbcConfig.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.commons.dbcp2.PoolingDataSource;
+import org.apache.commons.pool2.ObjectPool;
+import org.apache.commons.pool2.impl.GenericObjectPool;
+import org.springframework.beans.factory.config.ConfigurableBeanFactory;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
+
+@Configuration
```
jaycarlton: Do I really need a  configuration class for each properties class? I can put more than one of them in the enable annotation.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
jaycarlton: It almost seems like the client needs to build stairway. I can put boolean properties in for each hook you might want to add (assuming the full list is finite), but yeah the bean bag is another ball of wax. 🤔

src/main/java/bio/terra/common/stairway/StairwayComponent.java
jaycarlton: yeah this line was copied. I put in `namespace-suffix` for that piece.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
jaycarlton: @yonghaoy ^

src/main/java/bio/terra/common/stairway/StairwayComponent.java
jaycarlton: This was in the existing code.

src/main/java/bio/terra/common/db/JdbcConfig.java
jaycarlton: Oh interesting. I could be out of the database business except for the stairway db. I like that.

src/main/java/bio/terra/common/kubernetes/KubeConfig.java
jaycarlton: Yeah, the `@EnableConfigurationProperties(klass.class)` promotes the POJO to a bean. 🙃

jaycarlton:
regarding this [commment](https://github.com/DataBiosphere/terra-common-lib/pull/31#discussion_r593456010) on customizing the stairway construction, that makes sense, but it means Stairway can't be a bean. I guess that's not a huge loss, but it is a sore thumb.

src/main/java/bio/terra/common/stairway/StairwayConfig.java
jaycarlton: Distinguishing this from other strings when injecting it.

src/main/java/bio/terra/common/db/JdbcConfiguration.java
yonghaoy: @ddietterich But we are using dbcp2 today, e.g. https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/app/configuration/BaseDatabaseConfiguration.java

If this class would be the common base class for all apps, the easiest path to migrate is to keep using dbcp2 with jmx config for now, then switch to HikariCP in the future.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
jaycarlton: We actually can't make `Stairway` into a bean if we want to configure hooks and the context in the initialize method. This also means that the KubePodListener can't be provided as a bean.

ddietterich: Why does Stairway need to be a bean? Injection is a convenience, not a requirement.

I don't think KubePodListener should be bean. It is only used by KubeService, so never needs to be injected.

src/main/java/bio/terra/common/stairway/StairwayConfig.java
yonghaoy: This comes from https://github.com/DataBiosphere/terra-common-lib/pull/31#discussion_r594814785 (config and property)

src/main/java/bio/terra/common/stairway/StairwayComponent.java
jaycarlton: Good point. It could be an inner class, but it feels like too much code for that.

build.gradle
```diff
@@ -38,7 +38,9 @@ dependencies {
     api group: 'io.opencensus', name: 'opencensus-api', version: OPENCENSUS_VERSION
     api group: "io.opencensus", name: "opencensus-contrib-spring", version: OPENCENSUS_VERSION
     implementation group: "io.opencensus", name: "opencensus-exporter-trace-stackdriver", version: OPENCENSUS_VERSION
-    // commons-lang3 version is managed by the Spring Boot plugin
+
+    // commons-dbcp2 and -lang3 versions are managed by the Spring Boot plugin
+    implementation group: 'org.apache.commons', name: 'commons-dbcp2'
```
ddietterich, line 43: This line is a dup

src/main/java/bio/terra/common/kubernetes/KubeProperties.java
```diff
@@ -0,0 +1,52 @@
+package bio.terra.common.kubernetes;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "terra.common.kubernetes")
+public class KubeProperties {
+  // The pod that host the Buffer service app.
+  private String podNameFilter;
+
+  // Name of the Kubernetes pods we are running in. We uses common library uses this to list all
```
ddietterich: nit: should these be javadoc format?

src/main/java/bio/terra/common/stairway/StairwayComponent.java
ddietterich: I don't love it, but I see how it is used in Buffer Service, so nevermind.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -0,0 +1,142 @@
...
+            .enableWorkQueue(kubeProperties.isInKubernetes());
+    hooks.forEach(builder::stairwayHook);
+    try {
+      this.stairway = builder.build();
+    } catch (StairwayExecutionException e) {
+      throw new IllegalArgumentException("Failed to build Stairway.", e);
+    }
+
+    try {
+      // TODO(PF-161): Determine if Stairway and buffer database migrations need to be coordinated.
```
ddietterich: Comment is obsolete. PF-161 is done.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
ddietterich: This is certainly not an illegal argument. I think this is here because StairwayExecutionException was turned into a checked exception. We should probably undo using checked exceptions from Stairway. It has turned out to be a real PITA.

ddietterich (approved):
Minor comments.
LGTM

src/main/java/bio/terra/common/kubernetes/KubeConfig.java
```diff
@@ -0,0 +1,21 @@
...
+    return new KubeService(
+        kubernetesProperties.getPodName(),
```
yonghaoy: nit, do we want to use kubeProperites or kubenetesProperies in all places. Better to choose one instead of using both

src/main/java/bio/terra/common/stairway/StairwayComponent.java
yonghaoy: This class was designed for updating/tracking Stairway status as well. Also including the`showdown()` method that can switch status to SHUTDOWN, then buffer's service status will also get updated.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * Build and initialize the Stairway object
+   *
+   * @param dataSource data source for the Stairway DB
+   * @param context application context or other service-specific contextual object
+   * @param hooks list of Stairway hooks to install when building Stairway
+   */
+  public void initialize(DataSource dataSource, Object context, List<StairwayHook> hooks) {
+    logger.info("Initializing Stairway...");
+    //    logger.info("stairway username {}", stairwayProperties.getUsername());
```
yonghaoy: add it back or remove it?

src/main/java/bio/terra/common/stairway/StairwayComponent.java
yonghaoy: For now, I think RunTimeException makes more sense here until StairwayException becomes unchecked.

yonghaoy (approved):

src/main/java/bio/terra/common/stairway/StairwayComponent.java
jaycarlton: Had to remove it b/c that properties object doesn't have the username. It comes in with the data source.

PR #32, opened 2021-03-18
[PF-450] Adjust the logback context reset action for human-readable logging

gjuggler:
Context: I ran into a subtle issue when migrating RBS to the common logging module. Everything works fine when running the JSON format, but for the human-readable format, I wasn't seeing debug-level logs even though our application config was set to turn on debug logs for CRL:

```
logging.level.bio.terra.cloudres: DEBUG
```

A bit of debugging highlighted this stanza in the logging initializer. I'd been too aggressive in clearing out the logback context here, and the replacement (`logbackLogger.detachAndStopAllAppenders();`) seems to work better, at least for the issue I'd been seeing in RBS.

This isn't blocking anything on the JSON-logging for RBS. It's just a small quality of life improvement for human-readable logging.

### DIFF
--- a/src/main/java/bio/terra/common/logging/LoggingUtils.java
+++ b/src/main/java/bio/terra/common/logging/LoggingUtils.java
@@ -84,11 +84,18 @@ public class LoggingUtils {
     ConfigurableEnvironment environment = applicationContext.getEnvironment();
 
     if (Arrays.stream(environment.getActiveProfiles()).anyMatch("human-readable-logging"::equals)) {
-      System.out.println("Human-readable logging enabled, skipping Google JSON layout");
-      // Attempt to fully reload the logback logger context to reset to the default human-readable
-      // logback.xml config.
+      System.out.println("Human-readable logging enabled, re-applying original logback.xml config");
       try {
-        logbackLogger.getLoggerContext().reset();
+        // Note: there is some nuance in how best to reset the logback context. This code path is
+        // only encountered in (1) unit tests, where we actively need to reset the context since
+        // prior tests may have enabled the JSON layout, and (2) local testing of a service, where
+        // we want to disrupt as little of the original logging config as possible (e.g. Spring-
+        // configured log levels via "logging.level.foo.bar=DEBUG".
+        //
+        // Empirically, detaching all appenders and reloading the context via this logback file
+        // seems to work. But if we encounter future issues in human-readable logging control, this
+        // is a reasonable place to look more closely.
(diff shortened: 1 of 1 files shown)


gjuggler:
Tagging @marikomedlock as primary on this one since I know you've dived into logback stuff very recently.

jmczerk (approved):
Good docs should this come up again, LGTM

src/main/java/bio/terra/common/logging/LoggingUtils.java
```diff
@@ -84,11 +84,18 @@ protected static void initializeLogging(ConfigurableApplicationContext applicati
...
+        // Note: there is some nuance in how best to reset the logback context. This code path is
+        // only encountered in (1) unit tests, where we actively need to reset the context since
+        // prior tests may have enabled the JSON layout, and (2) local testing of a service, where
+        // we want to disrupt as little of the original logging config as possible (e.g. Spring-
+        // configured log levels via "logging.level.foo.bar=DEBUG".
+        //
+        // Empirically, detaching all appenders and reloading the context via this logback file
+        // seems to work. But if we encounter future issues in human-readable logging control, this
+        // is a reasonable place to look more closely.
+        logbackLogger.detachAndStopAllAppenders();
```
marikomedlock, line 98: RE "empirically": I agree wholeheartedly with this. Programmatically configuring logback is not very well documented and has been mostly trial and error on my part.

RE code path (1): Does this also work if there were child loggers? e.g. 1st unit test logs something using a class-specific (i.e. not root) logger, 2nd test changes the log format and again logs something using a class-specific logger.

When I add a CLI command to modify the log level, I'll basically have this same situation (log level changes mid-process). And I'm not sure whether just updating the appenders on the root logger is good enough, or whether I'll also have to update all existing child loggers (e.g. loggerContext.getLoggerList()). This might be fine, but I'm just not sure whether updating the root logger has recursive effects or not.

marikomedlock (approved):

src/main/java/bio/terra/common/logging/LoggingUtils.java
gjuggler: Good question re: root logger and inheritance. The unit tests are all using statically loaded class-specific loggers like `LoggerFactory.getLogger(LoggingTestController.class);` and they seem to pick up these changes fine. So it suggests to me that there's some amount of inheritance based on the logging context and the CLI use case should work OK... but the real proof is in the pudding of course.

PR #33, opened 2021-03-18
[no ticket] Add docs on locally testing through mavenLocal

gjuggler:
This makes it super easy to cross-check changes against a local service.

(We should probably add similar docs to the CRL readme too.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,22 +52,19 @@ psql -f local-dev/local-postgres-init.sql
 
 ### Local testing
 When working on a TCL package, it is often helpful to be able to quickly test out changes
-in the context of a service repo (e.g. terra-workspace-manager or terra-resource-buffer)
+in the context of a service repo (e.g. `terra-workspace-manager` or `terra-resource-buffer`)
 running a local server.
 
-The recommended way to iterate on local testing is by publishing TCL to a local Maven repo
-and loading the package from within the service repo:
+Gradle makes this very easy with a `mavenLocal` target for publishing and loading packages:
 
-1. Bump the TCL version number locally:
+1. Publish from TCL to your machine's local Maven cache.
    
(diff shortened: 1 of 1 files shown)


yonghaoy (approved):
Thanks! I usually just `gralde build` a local jar and copy that jar's absolute path. In service's build.gradle, I use implement(file(abs-path)). 
Yesterday, I saw what you did for buffer using new CRL package, and tried the new way.

gjuggler:
> Thanks! I usually just `gralde build` a local jar and copy that jar's absolute path. In service's build.gradle, I use implement(file(abs-path)).
> Yesterday, I saw what you did for buffer using new CRL package, and tried the new way.

Yup – I'd been doing this too. It can get annoying when the library changes involve adding new dependencies (I always had to manually add new deps in the service repo too). Using the `mavenLocal` approach handles new deps without any issue. Turns out the Gradle folks have thought this all through before 😂

README.md
```diff
@@ -45,9 +44,37 @@ PGPASSWORD=tclstairwaypwd psql postgresql://127.0.0.1:5432/tclstairway -U tclsta
...
+   
+   repositories {
+     mavenLocal()
+     mavenCentral()
+     ...
+   }
+   ```
+
+That's it! Your service should pick up locally-published changes. If your changes involved bumping 
+a minor version of a TCL package, be careful to update version numbers accordingly.
```
jaycarlton, line 80: How do I un-publish old versions?

jaycarlton (approved):

README.md
gjuggler: I think the official process is "Google a bunch until I find a stackoverflow example that seems about right" 😂 

More seriously though, it might be a nice addendum if you do find a clear path. To some extent, since we're only transiently adding `mavenLocal()` to the service repo, it shouldn't be too much of a concern.

PR #34, opened 2021-03-18
[no ticket] Use broadbot-specific secret for version bumps

gjuggler:
This follows Mariko's fix from https://github.com/DataBiosphere/terra-cli/pull/34/files. We need to use the Broadbot-specific token rather than the default repo token ([ref](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret)).

This would ideally be an org-wide secret to reduce manual setup churn; see DDO-1146 tracking that cleanup.


yonghaoy (approved):

marikomedlock (approved):
LGTM

PR #35, opened 2021-03-18
[PF-590]: Implement DatabaseConfiguration which config DB data source

yonghaoy:
Background: StairwayCompoment which manages Stairway life cycle can be moved to common library, but that needs Stairway DataSource. 
Also the whole logic is a common class. In the future, if client want to customize other connection pool, we can expose that like what we did for JMXEnable.

Buffer use:
https://github.com/DataBiosphere/terra-resource-buffer/pull/129


build.gradle
```diff
@@ -22,7 +22,7 @@ repositories {
 }
 
 group 'bio.terra'
-version '0.0.22-SNAPSHOT'
+version '0.0.22.2-SNAPSHOT'
```
ddietterich: ?? Is changing the version format part of this

src/main/java/bio/terra/common/db/DatabaseConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public DatabaseConfiguration(DatabaseProperties databaseProperties) {
+    this.databaseProperties = databaseProperties;
+  }
+
+  public DatabaseProperties getDatabaseProperties() {
+    return databaseProperties;
+  }
+  // Main use of the configuration is this pooling data source object.
+  public PoolingDataSource<PoolableConnection> getDataSource() {
```
ddietterich: Can this return DataSource, so we are not exposing the implementation detail?

src/main/java/bio/terra/common/db/DatabaseProperties.java
```diff
@@ -0,0 +1,87 @@
...
+
+  public void setUsername(String username) {
+    this.username = username;
+  }
+
+  public void setPassword(String password) {
+    this.password = password;
+  }
+
+  public boolean isInitializeOnStart() {
```
yonghaoy: Seems both Stariway DB and service DB use the same two boolean flag, and those seems to be a common config: 
`initialize` and `upgrade`

src/main/java/bio/terra/common/db/DatabaseConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+    // Lazy allocation of the data source
+    if (dataSource == null) {
```
ddietterich: I wonder if this is useful. In all known cases, we use the dataSource in everything we do, so being lazy has no benefit.

src/main/java/bio/terra/common/db/DatabaseProperties.java
```diff
@@ -0,0 +1,87 @@
...
+  public boolean isUpgradeOnStart() {
+    return upgradeOnStart;
+  }
+
+  public void setUpgradeOnStart(boolean upgradeOnStart) {
+    this.upgradeOnStart = upgradeOnStart;
+  }
+
+  @Override
+  public String toString() {
```
yonghaoy: not sure if we need those two, and I didn't find anywhere that we are using that.

src/main/java/bio/terra/common/db/DatabaseProperties.java
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * Returns a boolean indicating whether JMX should be enabled when creating connection pools. If
+   * this is enabled, references to connection pools created at context initialization time will be
+   * placed into an MBean scoped for the lifetime of the JVM. This is OK in production, but can be
+   * an issue for tests that use @DirtiesContext and thus repeatedly initialize connection pools, as
+   * these pools will hold active DB connections while held by JMX. Having this configuration allows
+   * us to disable JMX in test environments. See PF-485 for more details.
+   */
+  public boolean isJmxEnabled() {
```
yonghaoy: This is copied from @jmczerk 's Janitor code

src/main/java/bio/terra/common/db/DatabaseProperties.java
```diff
@@ -0,0 +1,87 @@
...
+    return upgradeOnStart;
+  }
+
+  public void setUpgradeOnStart(boolean upgradeOnStart) {
+    this.upgradeOnStart = upgradeOnStart;
+  }
+
+  @Override
+  public String toString() {
+    return new ToStringBuilder(this, ToStringStyle.JSON_STYLE)
```
ddietterich: Looks incomplete

yonghaoy (commented):
I just move StiarwayDbConfig here. Not sure if we need to do that for service DB. Those code are similar through even just few line(empoty class extened this).
If we should, I can create a `PrimaryDatabaseConfiguration` or `ServerPrimaryDatabaseConfiguration`, and replace `BufferDatabaseConfiguration` with this

src/main/java/bio/terra/common/db/DatabaseProperties.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+
+/** Defines configuration properties for using {@link DatabaseConfiguration} use. */
+public class DatabaseProperties {
+  private boolean jmxEnabled = true;
+  private String uri;
+  private String username;
+  private String password;
+
```
ddietterich: TDR has additional parameters it uses to control the size of its connection pool. I think everyone should be specifying these. Otherwise, you run the database server out of connections and get odd DB errors instead of pending on the connection pool.

See [DataRepo JdbcConfiguration](https://github.com/DataBiosphere/jade-data-repo/blob/abf95af809795b7166b431e08138094bfea39954/src/main/java/bio/terra/app/configuration/JdbcConfiguration.java#L24)

ddietterich (commented):
That looks promising. I note additional parameters are needed to properly configure the connection pool.

It will be worth documenting how a service would derive from these to make its own database data source.

src/main/java/bio/terra/common/db/DatabaseConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+import org.apache.commons.pool2.ObjectPool;
+import org.apache.commons.pool2.impl.GenericObjectPool;
+import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+
+/**
+ * Base class to config JDBC data source. Client can entend along with customized {@link DatabaseProperties} to
+ * build their data source. See {@link bio.terra.common.stairway.StairwayDatabaseConfiguration} to understand how to
+ * use in SpringBoot.
+ */
+public class DatabaseConfiguration {
```
jaycarlton: It would help to avoid naming classes configuration unless they were Spring `@Configuration` classes. Maybe name this `BaseDatabaseSettings` and make it abstract (or give it a protected ctor)?

build.gradle
yonghaoy: that was for my local testing. I modify the version here and `./gradlew publishToMavenLocal`. Then I can test the jar in my buffer service code. I will revert back once code LG to all reviewers

src/main/java/bio/terra/common/db/DatabaseProperties.java
jaycarlton: +1 I wasn't sure what to do with this either.

src/main/java/bio/terra/common/stairway/StairwayDatabaseConfiguration.java
```diff
@@ -0,0 +1,13 @@
+package bio.terra.common.stairway;
+
+import bio.terra.common.db.DatabaseConfiguration;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties(value = StairwayDatabaseProperties.class)
+public class StairwayDatabaseConfiguration extends DatabaseConfiguration {
+  public StairwayDatabaseConfiguration(StairwayDatabaseProperties StairwayJdbcProperties) {
```
jaycarlton: rename StairwayJdbcProperties

src/main/java/bio/terra/common/db/DatabaseConfiguration.java
yonghaoy: Moved initialization to constructer now

yonghaoy:
> That looks promising. I note additional parameters are needed to properly configure the connection pool.
> 
> It will be worth documenting how a service would derive from these to make its own database data source.

Thanks, added those two pool configs TDR has. Also in DatabaseConfiguraiton's javadoc, I said: 
see: `bio.terra.common.stairway.StairwayDatabaseConfiguration` to understand how to use it in SpringBoot. Is that enough for client understanding how to use it?

src/main/java/bio/terra/common/stairway/StairwayDatabaseProperties.java
```diff
@@ -0,0 +1,7 @@
+package bio.terra.common.stairway;
+
+import bio.terra.common.db.DatabaseProperties;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "terra.common.stairway.db")
```
jaycarlton: Oh good, so this trick works. 👍🏻

src/main/java/bio/terra/common/db/DatabaseProperties.java
```diff
@@ -0,0 +1,116 @@
+package bio.terra.common.db;
+
+import static org.apache.commons.pool2.impl.GenericObjectPoolConfig.DEFAULT_MAX_TOTAL;
+
+import org.apache.commons.dbcp2.*;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+
+/** Defines configuration properties for using {@link DatabaseConfiguration} use. */
+public class DatabaseProperties {
```
jaycarlton: define this as abstract if you want it to be used by subclasses only. Maybe name it `BaseDatabaseProperties`.

jaycarlton (approved):
I think this is just what I need. Thanks!

src/main/java/bio/terra/common/db/DatabaseConfiguration.java
yonghaoy: They are "@Configuration". All sub-classes has `@Configuration`.  And I think the "Configuration" name make sense here(it just puts all actual inside a proeprites class). And the configuration + properity pattern seems to be the common pattern we choose now

src/main/java/bio/terra/common/db/DatabaseProperties.java
yonghaoy: This can be used as a non-abstract class in a non-DI world. But there is no Terra service use it in that way. 
Not sure if we want to make it abstract and subclass-only or keep it in that way. 
@ddietterich wdyt? you implement the 1st version then we keep using this pattern since then.

yonghaoy: updated, add a abstract base and a PrimaryDbConfig. Does that LG?

yonghaoy: change my mind after sync after standup.

ddietterich (approved):
Seems very straightforward. Thanks

ddietterich (approved):
lgtm

PR #37, opened 2021-03-25
use cluster name suffix

jaycarlton:
The cluster name must be dynamic, so restoring the logic to  base it off the namespace and a service-specific suffix.

### DIFF
--- a/src/main/java/bio/terra/common/stairway/StairwayComponent.java
+++ b/src/main/java/bio/terra/common/stairway/StairwayComponent.java
@@ -139,6 +139,7 @@ public class StairwayComponent {
   }
 
   private String getClusterName() {
-    return String.format("%s-%s", kubeService.getNamespace(), stairwayProperties.getClusterNameSuffix());
+    return String.format(
+        "%s-%s", kubeService.getNamespace(), stairwayProperties.getClusterNameSuffix());
   }
 }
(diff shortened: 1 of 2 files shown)


yonghaoy (approved):

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -137,4 +137,8 @@ public Stairway get() {
   public StairwayComponent.Status getStatus() {
     return status.get();
   }
+
+  private String getClusterName() {
+    return String.format("%s-%s", kubeService.getNamespace(), stairwayProperties.getClusterNameSuffix());
+  }
```
marikomedlock: I don't think I have all the context on this change, but does this logic to construct a K8S cluster name by concatenating the namespace and a suffix apply to all services that use Stairway (e.g. will it apply to TDR also or only AP-style deployments)?

marikomedlock (approved):
Per discussion post-standup:
- Approving this change now to un-block progress on WSM multi-instance support.
- @jaycarlton is going to add a ticket to allow overriding the cluster name, so that this common library code can be used by services without assuming a particular devops setup (i.e. cluster name follows a particular pattern)

jaycarlton:
filed https://broadworkbench.atlassian.net/browse/PF-610 for follow-on work

PR #40, opened 2021-04-01
[PF-388] Add retry logic for Sam API calls to the common lib. 

tnadler:
Adapted from code in TDR. SamRetry always returns a SamApiException, that extracts the message from the Sam exception  and includes the HTTP Status code.


build.gradle
```diff
@@ -69,6 +69,8 @@ dependencies {
     testImplementation group: 'org.mockito', name: 'mockito-core', version: '2.+'
     testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-test'
     testImplementation group: 'org.openapitools', name: 'jackson-databind-nullable', version: '0.2.1'
+
+    compile 'org.broadinstitute.dsde.workbench:sam-client_2.12:0.1-343dfff-SNAP'
```
yonghaoy, line 74: this should be `implementation` in grade's new world. And maybe move. this close to other `implementation`?

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,188 @@
...
+
+import static java.time.Instant.now;
+
+/**
+ * SamRetry encapsulates logic needed for retrying Sam API calls. The retry function is linear and
+ * all constants are hard-coded, as Sam failures are generally just short periods of down-time.
+ */
+class SamRetry {
+    private static Logger logger = LoggerFactory.getLogger(SamRetry.class);
+    private static final int MAXIMUM_WAIT_SECONDS = 30;
```
yonghaoy: Suggest to use `Duration` for all time related fields.

yonghaoy (approved):
Overall LGTM. Thanks!

src/main/java/bio/terra/common/retry/SamRetry.java
wchamber: ditto

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,188 @@
...
+        void apply() throws ApiException, InterruptedException;
+    }
+
+    @FunctionalInterface
+    public interface SamFunction<R> {
+        R apply() throws ApiException, InterruptedException;
+    }
+
+
+    static <T> T retry(SamFunction<T> function)
```
wchamber: I think this should be public? ditto other static methods?

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,188 @@
...
+            } catch (ApiException ex) {
+                handleApiException(ex);
+            } catch (Exception ex) {
+                throw new InternalServerErrorException("Unexpected exception type: " + ex.toString(), ex);
+            }
+            sleepBeforeRetrying();
+        }
+    }
+
+    private void handleApiException(ApiException ex) {
```
wchamber: Hmmm, I'm not sure that we want to force users of the API to have their exceptions wrapped. Maybe offer two versions of retry functions, or offer the exception wrapping as a separate composable function? Not sure.

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,188 @@
...
+        ErrorReportException rex = convertSAMExToCommonEx(ex);
```
wchamber: nit: Avoid abbreviations. go/java-practices/naming#general-advice
ditto `fooEx` variants (other than in `catch` blocks`). Not so golang, Java is the land of the verbose.

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,188 @@
...
+        if (!(rex instanceof InternalServerErrorException)) {
+            throw rex;
+        }
+        logger.info("SamRetry: caught retry-able exception: " + ex);
```
wchamber: prefer the `.info(String, Exception)` overload

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,188 @@
...
+                throw new InternalServerErrorException("Unexpected exception type: " + ex.toString(), ex);
+            }
+            sleepBeforeRetrying();
+        }
+    }
+
+    private void performVoid(SamVoidFunction function) throws InterruptedException {
+        while (true) {
+            try {
+                function.apply();
```
wchamber: consider wrapping SamVoidFunction as a `SamFunction<Void>` and calling `perform` instead to cut down on duplicated code. e.g.
```
private void performVoid(SamVoidFunction function) throws InterruptedException {
  perform(() -> { function.apply(); return null; }); 
}

File: src/main/java/bio/terra/common/retry/SamRetry.java [hunk 8]
```diff
@@ -0,0 +1,188 @@
...
+            logger.error("SamRetry: operation timed out after " + operationTimeout.toString());
+            throw rex;
+        }
+    }
+
+    private void sleepBeforeRetrying() throws InterruptedException {
+        logger.info("SamRetry: sleeping " + retrySeconds + " seconds");
+        TimeUnit.SECONDS.sleep(retrySeconds);
+
+        retrySeconds = retrySeconds + retrySeconds;
```
wchamber: isn't this exponential backoff, double everytime? javadoc on class says linear.

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,188 @@
...
+        if (!(rex instanceof InternalServerErrorException)) {
+            throw rex;
+        }
+        logger.info("SamRetry: caught retry-able exception: " + ex);
+        // if our operation timeout will happen before we wake up from our next retry
+        // sleep, then we give up and re-throw.
+        if (operationTimeout.minusSeconds(retrySeconds).isBefore(now())) {
+            logger.error("SamRetry: operation timed out after " + operationTimeout.toString());
+            throw rex;
+        }
```
wchamber: optional style: it's a little surprising to me that this function checks the timeout and throws. consider making a private variable `lastException` and handle throwing if we're going to timeout in `sleepBeforeRetrying`. Or update the function name.

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,188 @@
...
+ * SamRetry encapsulates logic needed for retrying Sam API calls. The retry function is linear and
+ * all constants are hard-coded, as Sam failures are generally just short periods of down-time.
+ */
+public class SamRetry {
+    private static Logger logger = LoggerFactory.getLogger(SamRetry.class);
+    private static final int MAXIMUM_WAIT_SECONDS = 30;
+    private static final int INITIAL_WAIT_SECONDS = 10;
+    private static final int OPERATION_TIMEOUT_SECONDS = 300;
+    private final Instant operationTimeout;
+    private int retrySeconds;
```
wchamber: worth a comment on what this is

wchamber (changes requested):

build.gradle
tnadler: Done. Took the version from the WSM gradle file.

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Yes caught that in testing, thank you

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Good catch. TBH copied the retry function and am now wondering if the function as we have it here really makes sense. Updated the comment to reflect reality, but @ddietterich should we consider just having a linear retry of 30 seconds and forget the exponential bit?

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Done

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Done

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Done

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Done

src/main/java/bio/terra/common/retry/SamRetry.java
ddietterich: I think we should force a mapping based on exceptions from the common library that can be easily converted to an ErrorReport with the right HTTP status by a global exception handler.

At a minimum, we should convert it to an exception that has a message. The API messages from Sam are hard to use. Sometimes you have to dig into the next cause level to get any message text. There is code in TDR that does that.

src/main/java/bio/terra/common/retry/SamRetry.java
ddietterich: Nice!

src/main/java/bio/terra/common/retry/SamRetry.java
ddietterich: Our goal in TDR was to ride through long Sam outages caused by their merges into dev that could last for minutes. So exponential made sense. (We did several things to get our tests to run in the face of Sam merges...)

Linear retry sounds fine to me. If we need different behavior we can add it later when we know why we need it.

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Done, good idea!

src/main/java/bio/terra/common/retry/SamRetry.java
wchamber: Not trying to advocate for linear or exponential, just noting that the javadoc on this class promises linear but the impl looks exponential. Happy to go with whatever's been found in practice to be useful.

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Per offline convo with Will, going to leave these wrapped in the common exception classes. Adding more Sam specific exceptions did not seem to add much at this point

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Refactored

tnadler:
Tests now passing and code updated, PTAL @wchamber

build.gradle
```diff
@@ -23,7 +23,7 @@ repositories {
 }
 
 group 'bio.terra'
-version '0.0.37-SNAPSHOT'
+version '0.0.38-SNAPSHOT'
```
wchamber: nit: fwiw, I *think* this is automatically updated by a bot and doesn't need to be touched. but it's also harmless and will just get bumped again afaik.

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  public static <T> T retry(SamFunction<T> function) throws InterruptedException {
+    SamRetry samRetry = new SamRetry();
+    return samRetry.perform(function);
+  }
+
+  // Version of the retry function for testing so that we can override timeout
+  @VisibleForTesting
+  public static <T> T retry(SamFunction<T> function, Duration timeout) throws InterruptedException {
```
wchamber: nit: the `@VisibleForTesting` annotated methods can probably be package public instead of public if the tests are in the same package.

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,194 @@
...
+    if (retryDuration.compareTo(MAXIMUM_WAIT) > 0) {
+      retryDuration = MAXIMUM_WAIT;
+    }
+  }
+
+  /**
+   * Converts a SAM-specific ApiException to the common API exception, based on the HTTP status
+   * code.
+   */
+  private static ErrorReportException convertSAMExToCommonEx(final ApiException samEx) {
```
wchamber: nit: "Sam" isn't actually an acronym, so always lower case "am"

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,194 @@
...
+  private <T> T perform(SamFunction<T> function) throws InterruptedException {
+    while (true) {
+      try {
+        return function.apply();
+      } catch (ApiException ex) {
+        ErrorReportException errorReportException = convertSAMExToCommonEx(ex);
+        if (isRetryable(errorReportException)) {
+          logger.info("SamRetry: caught retry-able exception: ", ex);
+          sleepOrTimeoutBeforeRetrying(errorReportException);
+          continue;
```
wchamber: nit: `continue` unnecessary

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,194 @@
...
+   */
+  private static ErrorReportException convertSAMExToCommonEx(final ApiException samEx) {
+    logger.warn("SAM client exception code: {}", samEx.getCode());
+    logger.warn("SAM client exception message: {}", samEx.getMessage());
+    logger.warn("SAM client exception details: {}", samEx.getResponseBody());
+
+    // Sometimes the sam message is buried one level down inside of the error report object.
+    // If we find an empty message then we try to deserialize the error report and use that message.
+    String message = samEx.getMessage();
+    if (StringUtils.isEmpty(message)) {
```
wchamber: nit: `message.isEmpty()`

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,194 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.dsde.workbench.client.sam.ApiException;
+import org.broadinstitute.dsde.workbench.client.sam.model.ErrorReport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
+
+/**
+ * SamRetry encapsulates logic needed for retrying Sam API calls. All constants are hard-coded, as
+ * Sam failures are generally just short periods of down-time.
```
wchamber, line 16: Also worth noting that this parses/wraps the exceptions and why we think that's a good thing to do.

src/test/java/bio/terra/common/retry/SamRetryTest.java
```diff
@@ -0,0 +1,55 @@
...
+  private int count;
+
+  @Before
+  public void setup() {
+    count = 0;
+  }
+
+  @Test(expected = InternalServerErrorException.class)
+  public void testRetryTimeout() throws Exception {
+    SamRetry.retry(() -> testRetryFinishInner(100), Duration.ofSeconds(10));
```
wchamber: assertThrows is the preferred way to do exception assertions. https://howtodoinjava.com/junit5/expected-exception-example/
https://engdoc.corp.google.com/eng/doc/devguide/java/testing/exceptions.md?cl=head

wchamber (approved):

build.gradle
tnadler: Pretty sure I did not update this myself so maybe the bot already ran?

src/main/java/bio/terra/common/exception/SamApiException.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  public SamApiException(String message, ApiException apiException) {
+    super("Error from Sam: " + message, apiException);
+    this.apiException = apiException;
+  }
+
+  /** Get the HTTP status code of the underlying response from Sam. */
+  public HttpStatus getStatusCode() {
+    return HttpStatus.resolve(apiException.getCode());
```
wchamber: this can return null if it doesn't understand the status code. consider changing it to some default server error exception, or making the return value optional or null. I've seen the null happen on a request timeout.

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.concurrent.TimeUnit;
+import org.broadinstitute.dsde.workbench.client.sam.ApiException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
+
+/**
+ * SamRetry encapsulates logic needed for retrying Sam API calls. All constants are hard-coded, as
+ * Sam failures are generally just short periods of down-time.
+ * SamRetry always throws a SamApiException, which extracts the error message from Sam (which can be
```
wchamber: nit: use `<p>` for new paragraphs in javadoc

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  protected SamRetry(Duration timeout) {
+    this.operationTimeout = now().plus(timeout);
+    this.retryDuration = INITIAL_WAIT;
+  }
+
+  @FunctionalInterface
+  public interface SamVoidFunction {
+    void apply() throws ApiException, InterruptedException;
```
wchamber, line 47: do we need to allow `InterruptedException` in these interfaces? I don't think the underlying calls to SamService throw this.

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  /**
+   * Given an exception from Sam, either timeout and rethrow the error from Sam
+   * or sleep for retryDuration. If the thread times out while sleeping, throw the initial
+   * exception.
+   * @param samApiException The error Sam threw
+   * @throws SamApiException
+   */
+  private void sleepOrTimeoutBeforeRetrying(SamApiException samApiException)
```
wchamber: nit: maybe rename the argument to `previousException` or `initialException`?

src/main/java/bio/terra/common/exception/SamApiException.java
```diff
@@ -0,0 +1,86 @@
...
+ * Wrapper exception for non-200 responses from calls to Sam. Does not extend ErrorReportException
+ * so that SamApiException is a checked exception type. Can be converted to an ErrorReportException
+ * subtype using the toErrorReportException function.
+ * */
+public class SamApiException extends Exception {
+  private ApiException apiException;
+  private static final ObjectMapper objectMapper = new ObjectMapper();
+  private static Logger logger = LoggerFactory.getLogger(SamApiException.class);
+
+  static public SamApiException createSamApiException(ApiException apiException) {
```
wchamber: worth a few small unit tests on the conversion.

src/main/java/bio/terra/common/exception/SamApiException.java
```diff
@@ -0,0 +1,86 @@
...
+            objectMapper.readValue(apiException.getResponseBody(), ErrorReport.class);
+        message = errorReport.getMessage();
+      } catch (JsonProcessingException ex) {
+        logger.debug("Unable to deserialize sam exception response body");
+      }
+    }
+    return new SamApiException(message, apiException);
+  }
+
+  public SamApiException(String message, ApiException apiException) {
```
wchamber: Do we actually want this to be public, or should we always use the factory function?

wchamber (approved):
probably worth an update to the PR description

src/main/java/bio/terra/common/retry/SamRetry.java
ddietterich: The important thing is not to catch InterruptedExceptions while waiting. Those exceptions happen when Stairway needs to shut down its thread pool or otherwise cancel a flight.

If they never arise from Sam, then it isn't needed.

src/main/java/bio/terra/common/exception/SamApiException.java
tnadler: WDYT?

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: How would we verify that Sam doesn't throw these? Also is it possible that SamService in WSM doesn't, but somewhere else it might?

src/main/java/bio/terra/common/exception/SamApiException.java
tnadler: Done

src/main/java/bio/terra/common/exception/SamApiException.java
tnadler: Good point.

src/main/java/bio/terra/common/retry/SamRetry.java
ddietterich: Since InterruptedException is a checked exception, I don't think the code compiles if the underlying modules throw those exceptions.

wchamber: I don't think Sam will thorw these, but we can from our sleeps.
As-is, we're doing sleeping in our retry. If Stairway is trying to stop the flight, the interrupt will cause the sleep will to raise the InterruptedException, causing us to swallow the interrupt and throw the last seen Sam exception. This doesn't have any guarantees on what the step code does (continue working? throw a step retry?). It also seems bad if the Stairway code running the step expects to always receive the InterruptedException so it can take appropriate action on it.
Do we need to be really careful about InterruptedExceptions for Stairway?

ddietterich: Swallowing InterruptedException is a BadThing™
It stops us from gracefully shutting down flights.

wchamber: Good to know, I think we're doing that incorrectly in buffer in a couple places.

src/main/java/bio/terra/common/exception/SamApiException.java
wchamber: I think pick a default server error status code.

src/main/java/bio/terra/common/retry/SamRetry.java
yonghaoy: Thank @wchamber for pinging me this. Good to know. Now InterruptedException is heavily used in Buffer as it need to keep polling some GCP operations result. I will change buffer's this behavior

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Done.

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: If stairway stops a flight, would the interrupted exception thrown from the sleep call really have information useful to stairway? That it expects to get back? How would that work?

ddietterich: The interrupted exception is generated when Stairway asks its thread pool to terminate. The thread pool generates interrupts on all threads. The top level of the flight controller catches the InterruptedException. It disowns the flight, and if the queue is enabled, it queues the flight. It then returns, which returns the thread back to the thread pool. It doesn't need any extra information to do this.

Not all flights will see an InterruptedException. Only ones in wait states. Other flights may notice the request to shutdown at a step boundary where they check a shutting-down state flag and perform the same set of exit state changes.

The pool terminate times out at some point, and remaining flight threads will just be killed as the pod is shutdown. Those will have to be recovered by other pods or the restart of this pod.

tnadler: Okey dokey, InterruptedExceptions are being raised!

src/main/java/bio/terra/common/exception/SamApiException.java
```diff
@@ -0,0 +1,85 @@
...
+ * Wrapper exception for non-200 responses from calls to Sam. Does not extend ErrorReportException
+ * so that SamApiException is a checked exception type. Can be converted to an ErrorReportException
+ * subtype using the toErrorReportException function.
+ */
+public class SamApiException extends Exception {
+  private ApiException apiException;
+  private static final ObjectMapper objectMapper = new ObjectMapper();
+  private static Logger logger = LoggerFactory.getLogger(SamApiException.class);
+
+  public static SamApiException createSamApiException(ApiException apiException) {
```
wchamber: nit: `create` is a fine name for this. The `SamApiException` is implicit in the class name for the factory function. e.g. `SamApiException.createSamApiException` is repetitive.

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  protected SamRetry(Duration timeout) {
+    this.operationTimeout = now().plus(timeout);
+    this.retryDuration = INITIAL_WAIT;
+  }
+
+  @FunctionalInterface
+  public interface SamVoidFunction {
+    void apply() throws ApiException, InterruptedException;
```
wchamber, line 47: With all this said, I think the main thing is that `retry` throws an `InterruptedException`, whereas the SamFunction and SamVoidFunction interfaces don't need to be cause the underlying client library won't throw InterruptedExceptions. But leaving `throws InterruptedException` on the interface also doesn't really hurt anyone.

src/main/java/bio/terra/common/retry/SamRetry.java
```diff
@@ -0,0 +1,128 @@
...
+        return function.apply();
+      } catch (ApiException ex) {
+        if (isRetryable(ex)) {
+          logger.info("SamRetry: caught retry-able exception: ", ex);
+          sleepOrTimeoutBeforeRetrying(ex);
+        } else {
+          throw ex;
+        }
+      } catch (Exception ex) {
+        throw new InternalServerErrorException("Unexpected exception type: " + ex.toString(), ex);
```
wchamber: I'm not sure this is needed. But maybe atleast `catch (RuntimeException` so we don't accidentally lose an InterruptedException?
as-is, `try { return function.apply(); }` is allowed to throw an InterruptedException by the signature on the SamFunction interface, and this catch will catch and rethrow it as a InternalServerErrorException, which we're trying to avoid.

src/test/java/bio/terra/common/exception/SamApiExceptionTest.java
```diff
@@ -0,0 +1,105 @@
...
+    ErrorReport errorReport = new ErrorReport().message(message);
+    try {
+      String errorReportString = objectMapper.writeValueAsString(errorReport);
+      ApiException apiException =
+          new ApiException(
+              "", HttpStatusCodes.STATUS_CODE_BAD_REQUEST, new HashMap<>(), errorReportString);
+      SamApiException samApiException = SamApiException.createSamApiException(apiException);
+      assertEquals("Error from Sam: " + message, samApiException.getMessage());
+    } catch (JsonProcessingException e) {
+      // FAIL
```
wchamber: I don't think this fails. Fine to add `throws Exception` to an `@Test` method and Junit will handle failing the `@Test` if some exception is thrown.

src/test/java/bio/terra/common/retry/SamRetryTest.java
```diff
@@ -0,0 +1,71 @@
...
+  @BeforeEach
+  public void setup() {
+    count = 0;
+  }
+
+  @Test
+  public void testRetryTimeout() throws Exception {
+    assertThrows(
+        ApiException.class,
+        () -> SamRetry.retry(() -> testRetryFinishInner(100), Duration.ofSeconds(10)));
```
wchamber, line 32: Let's try to add one test where we throw an InterruptedException by starting the work on another thread and interrupting it while it's sleeping on its retry.

wchamber (approved):

src/main/java/bio/terra/common/exception/SamApiException.java
tnadler: Done

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Ah OK. Well I will leave it for now then.

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: yeah I think this is not needed at all

src/test/java/bio/terra/common/exception/SamApiExceptionTest.java
tnadler: Ah my bad

tnadler (commented):
Thanks! OK to submit?

wchamber:
> Thanks! OK to submit?

Fine by me. Don't know if Dan wants another pass.
per last review comment, I think a unit test for the InterruptedException being thrown if the Thread.sleep is interrupted would be good since that's something we're explicitly trying to make happen, but won't block the PR on it.

ddietterich:
> > Thanks! OK to submit?
> 
> Fine by me. Don't know if Dan wants another pass.
> per last review comment, I think a unit test for the InterruptedException being thrown if the Thread.sleep is interrupted would be good since that's something we're explicitly trying to make happen, but won't block the PR on it.

I tried to take another look, but I am not seeing the changes that are reported in the comments as "Done". Maybe you need to push again?

tnadler:
@ddietterich which change? It looks up to date but I could easily have messed something up in a merge...

src/main/java/bio/terra/common/exception/SamApiException.java
ddietterich: It does not look like you changed the method name.

src/main/java/bio/terra/common/retry/SamRetry.java
ddietterich: If it is not needed, it should be removed. If you leave it, you should change it to `RuntimeException` so it will not catch `InterruptedException`

src/test/java/bio/terra/common/exception/SamApiExceptionTest.java
ddietterich: So I think you should remove the catch of JsonProcessingException, yes?

src/test/java/bio/terra/common/retry/SamRetryTest.java
ddietterich: +1

ddietterich (commented):
I commented on the spots where I thought you would change things, but nothing appears to be changed - at least in my github view...

src/test/java/bio/terra/common/retry/SamRetryTest.java
tnadler: Added testInterruptedExceptionThrown below

src/main/java/bio/terra/common/exception/SamApiException.java
tnadler: Shows up for me as changed now...

src/main/java/bio/terra/common/retry/SamRetry.java
tnadler: Should be gone now.

src/test/java/bio/terra/common/exception/SamApiExceptionTest.java
tnadler: Should be gone now

tnadler (commented):
Apologies for the git mess up. *Should* all be updated now, and added that test.

ddietterich (approved):
LGTM - thanks, Tova

PR #41, opened 2021-04-02
Demo protobufs for FlightMap serialization. [not for review]

wchamber:
Show how we could use protobufs for serializing data to store within FlightMaps, with clearly defined schema evolution properties.
See https://developers.google.com/protocol-buffers/docs/proto3

### DIFF
--- a/src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
+++ b/src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
@@ -1,8 +1,16 @@
 package bio.terra.common.iam;
 
+import bio.terra.common.exception.InternalServerErrorException;
 import bio.terra.common.exception.UnauthorizedException;
+import bio.terra.common.iam.proto.AuthenticatedUserRequestModel;
+import bio.terra.stairway.FlightMap;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.Message;
+import com.google.protobuf.Parser;
 import java.util.Objects;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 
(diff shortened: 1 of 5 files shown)


src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
```diff
@@ -119,4 +125,69 @@ public void testVectors() throws JsonProcessingException {
...
+            .setSubjectId(SUBJECT_ID)
+            .setRenamedToken(TOKEN)
+            .setAdditionalField("foo")
+            .build();
+
+    FlightMap inMap = new FlightMap();
+    putMessage(inMap, "v1", messageV1);
+    putMessage(inMap, "v2", messageV2);
+    FlightMap outMap = new FlightMap();
+    outMap.fromJson(inMap.toJson());
```
jmczerk, line 174: I'd be interested to see the pretty-printed output of `fromJson()` to assess human readability for debug.

src/main/proto/bio/terra/common/iam/iam.proto
```diff
@@ -0,0 +1,23 @@
...
+  optional string email = 1;
+  optional string subject_id = 2;
+  optional string token = 3;
+}
+
+// DON'T DO THIS
+// This is not the correct way to change a proto schema - you would always modify the message in
+// place, e.g. add the 'additional_field'. For illustrative purposes only, we define two schemas to
+// be compiled into the same binary to show forward/backwards compatibility.
+message AuthenticatedUserRequestModelV2 {
```
yonghaoy, line 17: I have never try clients depends on proto file. Is `bio.terra.common.iam.proto.AuthenticatedUserRequestModel` would automattly built into published package? (I am thinking about the pain when we have to upload/download yaml file)

src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
```diff
@@ -40,6 +48,58 @@ public String getToken() {
...
+        getMessage(flightMap, key, AuthenticatedUserRequestModel.parser());
+    return fromModel(model);
+  }
+
+  /**
+   * This could be put on FlightMap as a part of Stairway. Broken out to a separate function here to
+   * illustrate.
+   */
+  @VisibleForTesting
+  static FlightMap putMessage(FlightMap flightMap, String key, Message message) {
```
yonghaoy, line 66: Worth to add explain how 
https://github.com/DataBiosphere/stairway/blob/develop/src/main/java/bio/terra/stairway/FlightMap.java#L33
 and https://github.com/DataBiosphere/stairway/blob/develop/src/main/java/bio/terra/stairway/FlightMap.java#L51
would work in this case?

src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
wchamber: The binary format is not aimed at human readability: https://developers.google.com/protocol-buffers/docs/encoding
protobufs do have a pretty print text format encoding (& a JSON encoding), but they have different schema evolution properties (e.g. the field name is used, instead of the field tag number).
Adding some logger.info with comments for visibility.

src/main/proto/bio/terra/common/iam/iam.proto
wchamber: The generated proto java classes are included in the jar the same as any other class.
With regards to sharing .proto files in a better way than we're doing with the upload/download yaml for open api schema, you can depend on the .proto files of a dependency directly in addition to just it's published classes. This makes it possible to share & version control .proto files as a normal part of package dependencies. So, no more pain, I think, but haven't tried for myself.
See https://github.com/google/protobuf-gradle-plugin#protos-in-dependencies

To be clear, the .proto files used only for stairway serialization shouldn't need to be shared. It's fine to use the java generated classes, or just leave it to the implementation of the shared class (`AuthenticatedUserRequest`).

src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
wchamber: As those are package private, I can't demo them in this PR :(
Theoretically, they work the same as they do for any String value stored in a FlightMap. As-is in the demo, we're storing the binary serialized proto wire format as a String; it doesn't matter to Stairway that the String has special meaning.

PR #42, opened 2021-04-05
[PF-652]: Retry when getting pod list

yonghaoy:
I didn't spend too much effort on generalizing the retry part, although all API calls looks similar, but that is still special as Terra client library may have that covered(I over heard). For now, just unblock RBS release.

### DIFF
--- a/src/main/java/bio/terra/common/kubernetes/KubeService.java
+++ b/src/main/java/bio/terra/common/kubernetes/KubeService.java
@@ -103,7 +103,7 @@ public class KubeService {
       try {
         CoreV1Api api = makeCoreApi();
         V1PodList list =
-                api.listNamespacedPod(namespace, null, null, null, null, null, null, null, null, null);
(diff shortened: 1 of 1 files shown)


jmczerk (approved):

src/main/java/bio/terra/common/kubernetes/KubeService.java
```diff
@@ -92,22 +98,32 @@ public KubeService(String podName, boolean inKubernetes, String podNameFilter) {
...
+      try {
+        TimeUnit.MILLISECONDS.sleep(RETRY_SLEEP.toMillis());
+      } catch (InterruptedException e) {
+        throw new KubeApiException("Error listing pods", e);
       }
-      return pods;
-    } catch (ApiException ex) {
-      throw new KubeApiException("Error listing pods", ex);
     }
+    throw new KubeApiException("Unable to get Podlist");
```
wchamber: it'd be nice to include the last ApiException seen in this exception

wchamber (approved):

PR #43, opened 2021-04-06
Demo Avro for FlightMap serialization. [not for review]

jmczerk:
Avro equivalent of @wchamber protobuf demo.  

Avro schema evolution rules: https://avro.apache.org/docs/1.10.2/spec.html#Schema+Resolution

### DIFF
--- a/src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
+++ b/src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
@@ -1,9 +1,27 @@
 package bio.terra.common.iam;
 
+import bio.terra.common.exception.InternalServerErrorException;
 import bio.terra.common.exception.UnauthorizedException;
+import bio.terra.common.iam.avro.AuthenticatedUserRequestModel;
+import bio.terra.stairway.FlightMap;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import com.google.common.annotations.VisibleForTesting;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Objects;
+import org.apache.avro.Schema;
+import org.apache.avro.io.DatumReader;
+import org.apache.avro.io.Decoder;
+import org.apache.avro.io.DecoderFactory;
+import org.apache.avro.io.Encoder;
+import org.apache.avro.io.EncoderFactory;
+import org.apache.avro.specific.SpecificDatumReader;
+import org.apache.avro.specific.SpecificDatumWriter;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 
 /** Class representing the identity of an authenticated user. */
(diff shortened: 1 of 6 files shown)


src/test/java/bio/terra/common/iam/AuthenticatedUserRequestTest.java
```diff
@@ -119,4 +125,108 @@ public void testVectors() throws JsonProcessingException {
...
+    AuthenticatedUserRequest request =
+        AuthenticatedUserRequest.builder()
+            .setEmail(EMAIL_ADDRESS)
+            .setSubjectId(SUBJECT_ID)
+            .setToken(TOKEN)
+            .build();
+
+    FlightMap inMap = request.putIn(new FlightMap(), "foo");
+    FlightMap outMap = new FlightMap();
+    outMap.fromJson(inMap.toJson());
```
wchamber, line 140: we should both update to also use the test fixture for stairway serialization added in https://github.com/DataBiosphere/stairway/pull/64 to check for input FlightMap compatibility.

src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
```diff
@@ -40,6 +58,95 @@ public String getToken() {
...
+   * illustrate.
+   */
+  @VisibleForTesting
+  static <T> FlightMap putRecord(
+      FlightMap flightMap, String key, Schema schema, T record, Class<T> type) {
+    try {
+
+      // Encode the object with the "writer" schema into the byte stream.
+      OutputStream stream = new ByteArrayOutputStream();
+      Encoder jsonEncoder = EncoderFactory.get().jsonEncoder(schema, stream);
```
wchamber, line 91: any particular reason to use the json encoder instead of the binary encoder?

jmczerk: As one would expect, binary encoded data looks like a lot like protobuf:
```
 test@example.com\u000ESubject 0123.456-789AbCd\u0000\u0006foo
```

The big downside here is that "writer schema" must be stored with the data, and is JSON only; Avro is designed to be used either with large data sets where the schema is stored in a header with many records, or using a [schema registry](https://www.confluent.io/product/confluent-platform/data-compatibility) where a fingerprint is used as a key.  We could do our own schema management, but this seems like a lot of added complexity, and kind of moves backwards on the tight-coupling we're trying to get away from (assuming the schemas get stored in the database).

On the positives, I do like the no-code-generation option (see other PR); if we were to go with Avro with code generation I suspect there's a bit of work we'd want to do to tweak the [Velocity templates](https://velocity.apache.org/) that Avro uses to generate the classes

jmczerk: Btw, there may also be no-code-generation options for protobuf: https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/DynamicMessage

wchamber: Ha, I actually really dislike the no-code-generation options. Give me generated classes and type safety where I can't fat finger the field name.

jmczerk: In general I agree, but a bad object mode for the generated code can be a downside, which I think is reflected in the semantics of interaction with FlightMap here.

Protobuf seems to have been pretty thoughtfully designed their Java object model.  Avro can achieve something similar, but would require more customization to get there.  Out of the box the FlightMap-facing interface is awkward.

PR #46, opened 2021-04-06
Demo for Thrift serialization.

jmczerk:
So this was brutal, but here we go...

Biggest issue here is that the Gradle plugin still requires the Thrift compiler to be installed locally.  No precompiled MacOS bits were available outside homebrew, so I tried getting the Thrift compiler built from source... then ran into dependency issues with Bison, plus the requirement to install Boost and CMake ...  So I ended up hacking around all of this by pulling down a Ubuntu docker image and installing there with `apt`, and dropping a script to call into the container into `/usr/local/bin` ...
So there would be Gradle work beyond what's seen here that we'd have to do if we wanted to adopt this.


src/main/java/bio/terra/common/iam/AuthenticatedUserRequest.java
```diff
@@ -40,6 +50,61 @@ public String getToken() {
...
+    AuthenticatedUserRequestModel model =
+        getMessage(flightMap, key, new AuthenticatedUserRequestModel());
+    return AuthenticatedUserRequest.builder()
+        .setEmail(model.getEmail())
+        .setSubjectId(model.getSubjectId())
+        .setToken(model.getToken())
+        .build();
+  }
+
+  @VisibleForTesting
```
jmczerk, line 67: 🤢

src/main/thrift/iam.thrift
```diff
@@ -0,0 +1,16 @@
+namespace java bio.terra.common.iam.thrift
+
+struct AuthenticatedUserRequestModel {
```
yonghaoy, line 3: One interesting to see (and also pain using Swagger as Pojo) is polymorphism support
I didn't find a clear built-in support in thrift yet, is that simple?

jmczerk: By polymorphism do you mean having the ability to build messages off a common base message at the IDL level?  Thrift supports this for RPC services but not serialized data types (`struct`); generally this is approximated using composition.

Or do you mean polymorphism in the generated code (which it does, but it's messy; you can see in the signatures of the `putMessage()`/`getMessage()` methods in `AuthenticatedUserRequest`.

PR #48, opened 2021-04-16
[PF-666] Add common exceptions to the common lib.

tnadler:
Mostly copied from the WSM common exception directory, with a few comments cleaned up.


src/main/java/bio/terra/common/exception/InconsistentFieldsException.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.common.exception;
+
+public class InconsistentFieldsException extends BadRequestException {
```
ddietterich, line 7: I wouldn't have thought this one would be common, but OK.
We probably need better documentation for what these are supposed to mean.
No necessarily as part of this PR.

ddietterich (approved):
LGTM
We probably need better doc'n for these exceptions, but that shouldn't stop this PR.

src/main/java/bio/terra/common/exception/SamApiException.java
```diff
@@ -22,6 +22,10 @@
   private static Logger logger = LoggerFactory.getLogger(SamApiException.class);
 
   public static SamApiException create(ApiException apiException) {
+    return SamApiException.create("", apiException);
```
wchamber: nit: avoid using "magic values" like the empty string. Either use null or Optional empty if there's no valid messagePrefix.

src/main/java/bio/terra/common/exception/SamApiException.java
```diff
@@ -34,7 +38,11 @@ public static SamApiException create(ApiException apiException) {
         logger.debug("Unable to deserialize sam exception response body");
       }
     }
-    return new SamApiException(message, apiException);
+    if (!messagePrefix.isEmpty()) {
```
wchamber: nit: if we do StringUtils.isEmpty, we'll handle the null case.

src/main/java/bio/terra/common/exception/stairway/StairwayInitializationException.java
```diff
@@ -0,0 +1,18 @@
+package bio.terra.common.exception.stairway;
+
+import bio.terra.stairway.exception.StairwayRuntimeException;
+
+public class StairwayInitializationException extends StairwayRuntimeException {
```
wchamber: Seems kind of funny to be subclassing a stairway exception.

wchamber (approved):
Some of the non-ErrorReportException exceptions don't seem especially common, or particularly important to make common.
I think it'd be fine to leave them out and keep having exceptions defined in WM for its particular usage.
Unlike Dan, I think the time to add documentation is now :) If we can't explain the exception's use well, maybe a good signal to not include it.

tnadler:
That is fair. Will - when you say non-ErrorReport, do you mean those that don't derive directly from ErrorReport? Most of them derive from something that derives from ErrorReport. I do not have great context on these, was mostly just going off the common directory. Given this comment, I think I will revert InconsistentFieldsException. MDCHandlingException, and StairwayInitializationException. The others added seem more clearly common - MissingRequiredFieldException, NotImplementedException, SerializationException, ValidationException.

Thoughts before I go do that?

wchamber:
> That is fair. Will - when you say non-ErrorReport, do you mean those that don't derive directly from ErrorReport? Most of them derive from something that derives from ErrorReport. I do not have great context on these, was mostly just going off the common directory. Given this comment, I think I will revert InconsistentFieldsException. MDCHandlingException, and StairwayInitializationException. The others added seem more clearly common - MissingRequiredFieldException, NotImplementedException, SerializationException, ValidationException.
> 
> Thoughts before I go do that?

Reverting MDCHandlingException and StairwayInitializationException seem good to me.
I *think* InconsistentFieldsException is fairly analogous to MissingRequiredFieldException. (the fields are conflicting in a way that means the request is bad, but the fields aren't "missing" per se.) So that probably comes along.

tnadler:
SGTM. Confident enough in that assessment for me to put it in a comment? :)

wchamber:
> SGTM. Confident enough in that assessment for me to put it in a comment? :)

Sure.

src/main/java/bio/terra/common/exception/stairway/StairwayInitializationException.java
tnadler: reverted

src/main/java/bio/terra/common/exception/SamApiException.java
tnadler: done

src/main/java/bio/terra/common/exception/SamApiException.java
tnadler: done

src/main/java/bio/terra/common/exception/InconsistentFieldsException.java
tnadler: Updated comment and leaving here, per discussion.

ddietterich: My original intent for this exception was to mean we got an API request that is internally inconsistent. The one case is where the caller tries to assign a user to a private resource.

I see I used it a few other places that should probably be logic exceptions - 500s - not InconsistentFieldsExceptions.

ddietterich:
> Some of the non-ErrorReportException exceptions don't seem especially common, or particularly important to make common.
> I think it'd be fine to leave them out and keep having exceptions defined in WM for its particular usage.
> Unlike Dan, I think the time to add documentation is now :) If we can't explain the exception's use well, maybe a good signal to not include it.

I cannot disagree with that: If we don't know what it means, we shouldn't put it in common

PR #49, opened 2021-04-21
[PF-388] Rework Sam Exceptions. 

tnadler:
Create a family of Sam Exceptions derived from the corresponding ErrorReportExceptions so they are not checked. Move to a factory model for creating this exception type.


src/main/java/bio/terra/common/sam/exception/SamBadRequestException.java
```diff
@@ -0,0 +1,13 @@
+package bio.terra.common.sam.exception;
+
+import bio.terra.common.exception.BadRequestException;
+
+public class SamBadRequestException extends BadRequestException {
```
wchamber, line 9: nit: Worth some very short javadoc comment on what these Sam exceptions should be used for. ditto others.

src/main/java/bio/terra/common/sam/exception/SamExceptionFactory.java
```diff
@@ -0,0 +1,76 @@
...
+      }
+    }
+    if (!StringUtils.isEmpty(messagePrefix)) {
+      message = messagePrefix + ": " + message;
+    }
+    switch (apiException.getCode()) {
+      case HttpStatusCodes.STATUS_CODE_BAD_REQUEST:
+        {
+          return new SamBadRequestException(message, apiException);
+        }
```
wchamber: do we need the brackets for every case?

wchamber (approved):

ddietterich (approved):
LGTM

src/main/java/bio/terra/common/sam/exception/SamBadRequestException.java
tnadler: Done

src/main/java/bio/terra/common/sam/exception/SamExceptionFactory.java
tnadler: Done

PR #51, opened 2021-04-23
PF-671 - Introduce builder for initialize params; add exceptionSerializer

ddietterich:
The main goal is to plumb in the exception serializer for Stairway initialization that was lost in the move to common lib.

I decided to add a parameter builder for the StairwayComponent initializer method. I expect we will be adding more parameters here. The builder makes it simple to add more parameters over time. Let me know what you think.

### DIFF
--- a/src/main/java/bio/terra/common/stairway/StairwayComponent.java
+++ b/src/main/java/bio/terra/common/stairway/StairwayComponent.java
@@ -1,5 +1,7 @@
 package bio.terra.common.stairway;
 
+import static com.google.cloud.ServiceOptions.getDefaultProjectId;
+
 import bio.terra.common.kubernetes.KubeProperties;
 import bio.terra.common.kubernetes.KubeService;
 import bio.terra.stairway.ExceptionSerializer;
@@ -7,20 +9,17 @@ import bio.terra.stairway.Stairway;
 import bio.terra.stairway.StairwayHook;
 import bio.terra.stairway.exception.StairwayException;
 import bio.terra.stairway.exception.StairwayExecutionException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Component;
-
-import javax.sql.DataSource;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
-
-import static com.google.cloud.ServiceOptions.getDefaultProjectId;
+import javax.sql.DataSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
 
 /** A Spring Component for exposing an initialized {@link Stairway}. */
 @Component
--- a/src/main/java/bio/terra/common/stairway/StairwayComponent.java
+++ b/src/main/java/bio/terra/common/stairway/StairwayComponent.java
@@ -46,20 +46,21 @@ public class StairwayComponent {
         getClusterName());
   }
 
-  /** convenience for getting a parameter builder for initialize */
-  public InitializerBuilder newInitializeBuilder() {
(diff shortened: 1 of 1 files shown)


zloery (approved):
LGTM

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -142,4 +152,63 @@ private String getClusterName() {
...
+
+  /**
+   * Builder for providing inputs to the initialize method. This allows us to add parameters without
+   * adding new methods to the Stairway component.
+   */
+  public static class InitializerBuilder {
+    private DataSource dataSource;
+    private Object context;
+    private List<StairwayHook> hooks = new ArrayList<>();
+    private ExceptionSerializer exceptionSerializer;
```
wchamber, line 173: WDYT about having a `Function<Stairway.Builder, Stairway.Builder>` (or `Consumer<Stairway.Builder>`) to allow clients to mess with the underlying Stairway.Builder anyway they want? That means we don't need to keep extending this object for every new field on Stairway.Builder. Yes, it does allow the client to shoot themselves in the foot by overriding fields `StairwayComponent` sets, but I'm not convinced preventing bad ideas is statically enforceable anyway. We could always set the "StairwayComponent reserved fields" after applying the clients `Function`

We could either make that a field on `InitializerBuilder` or just get rid of the wrapping `InitializerBuilder` and take the `Function` directly in `initialize`. e.g.
```
// StairwayComponent.java
public void initialize(Function<Stairway.Builder, Stairway.Builder> stairwayCustomization);

// client
stairwayComponent.initialize(stairwayBuilder -> {
    return stairwayBuilder
      .applicationContext(myContext)
      .exceptionSerializer(mySerializer);
    // etc.
  });

File: src/main/java/bio/terra/common/stairway/StairwayComponent.java [hunk 2]
```diff
@@ -60,17 +59,28 @@ public StairwayComponent(
    * @param hooks list of Stairway hooks to install when building Stairway
    */
   public void initialize(DataSource dataSource, Object context, List<StairwayHook> hooks) {
```
wchamber, line 62: mark as `@Deprecated`?

src/main/java/bio/terra/common/stairway/StairwayComponent.java
ddietterich: One of the goals of the StairwayComponent was to have a consistent set of stairway configuration parameters that we could share. Sharing the setting of the parameters between StairwayComponent and service code seems confusing and hard to use to me. Yes it is a PITA to extend this builder when a new runtime Stairway parameter is added.

It won't be every time though. For example, the garbage collection controls are all config params and won't show up here. Only runtime things need to be provided on initialize.

I think the exception serializer can become common for us, if/when we move JobControl into the common lib.

In any case, my intent here was to solve this one problem for WSM production, not to redesign the StairwayComponent. If you don't want this introduced, I'll make a new method that passes in the serializer. This seemed more future-proof to me.

wchamber: SG, adding a builder seems better than a new method or continuing to extend the initialize parameters.
The parameter sharing does seem confusing, but agreed on punting a larger redesign.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -142,4 +152,63 @@ private String getClusterName() {
...
+    OK,
+    ERROR,
+    SHUTDOWN,
+  }
+
+  /**
+   * Builder for providing inputs to the initialize method. This allows us to add parameters without
+   * adding new methods to the Stairway component.
+   */
+  public static class InitializerBuilder {
```
wchamber: optional nit: maybe `InitializationCustomization` or `StairwayCustomization`? While this is a builder class relating to the initialize function, `InitializerBuilder` isn't conveying a lot to me.

wchamber (approved):

src/main/java/bio/terra/common/stairway/StairwayComponent.java
ddietterich: I renamed it `StairwayOptionsBuilder`

PR #52, opened 2021-04-26
[PF-666] Add new Exception type for when Sam service calls or retries are interrupted.

tnadler:
This Exception is unchecked and should be used when Sam is called outside of a Flight.

### DIFF
--- a/src/main/java/bio/terra/common/sam/exception/SamExceptionFactory.java
+++ b/src/main/java/bio/terra/common/sam/exception/SamExceptionFactory.java
@@ -59,4 +59,9 @@ public class SamExceptionFactory {
         return new SamInternalServerErrorException(message, apiException);
     }
   }
+
+  public static ErrorReportException create(
+      String messagePrefix, InterruptedException interruptedException) {
+    return new SamInterruptedException(messagePrefix, interruptedException);
+  }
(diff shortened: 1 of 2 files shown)


ddietterich (approved):
lgtm

src/main/java/bio/terra/common/sam/exception/SamExceptionFactory.java
```diff
@@ -59,4 +59,9 @@ public static ErrorReportException create(String messagePrefix, ApiException api
         return new SamInternalServerErrorException(message, apiException);
     }
   }
+
+  public static ErrorReportException create(
+      String messagePrefix, InterruptedException interruptedException) {
```
wchamber, line 69: nit: worth duplicating some of the comment on SamInterruptedException on when this method should be thrown, or just don't have it in this factory since there's nothing wrong with the constructor and we're not trying to hide something about creation.

wchamber (approved):

src/main/java/bio/terra/common/sam/exception/SamExceptionFactory.java
tnadler: Added comment.

PR #53, opened 2021-04-27
[no ticket]: fix bump mode

yonghaoy:
1. Every PR merge, bump minor version
2. When release, bump major version


ddietterich (changes requested):
I think this is not the right change.
I think it should be done as written here: [versioning proposal](https://docs.google.com/document/d/1eT-k48cxk74kScxP4iscg9lYETziK1zOivC1rnEpCzM/edit#heading=h.4govmyd845l5)

PR #56, opened 2021-05-21
jczerk/add workflow dispatch

jmczerk:
I would have liked to make the bump+publish trigger a little more full featured like [Workspace Manager's](https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/.github/workflows/tag-publish.yml#L30), but this is really a hedge against recent flakiness in GitHub actions.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -2,6 +2,7 @@ name: Publish package to Release Artifactory
 on:
   release:
     types: [created]
+  workflow_dispatch: {}
 env:
   VAULT_ADDR: https://clotho.broadinstitute.org:8200
   ARTIFACTORY_ACCOUNT_PATH: secret/dsp/accts/artifactory/dsdejenkins
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -2,7 +2,6 @@ name: Publish package to Release Artifactory
 on:
   release:
     types: [created]
-  workflow_dispatch: {}
 env:
   VAULT_ADDR: https://clotho.broadinstitute.org:8200
   ARTIFACTORY_ACCOUNT_PATH: secret/dsp/accts/artifactory/dsdejenkins
--- a/.github/workflows/develop-push.yml
+++ b/.github/workflows/develop-push.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - develop
+  workflow_dispatch: {}
 env:
   VAULT_ADDR: https://clotho.broadinstitute.org:8200
   ARTIFACTORY_ACCOUNT_PATH: secret/dsp/accts/artifactory/dsdejenkins
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,7 @@ on:
     branches: [ develop ]
   pull_request:
     branches: [ '**' ]
+  workflow_dispatch: {}
 jobs:
   test:
     runs-on: ubuntu-latest


yonghaoy (approved):

.github/workflows/publish-release.yml
```diff
@@ -2,6 +2,7 @@ name: Publish package to Release Artifactory
 on:
   release:
     types: [created]
+  workflow_dispatch: {}
```
yonghaoy: This is already manually triggered by creating a github release. In case of GHA hanging and we want to try, what would be the preferred process? Manually create a github release or direct launch this GHA. Feels former is better and more trackable?

yonghaoy (approved):

wchamber (approved):
So this is for the case when we can still run some GHA, but some are stuck? No objections, but I'm not especially informed either :)

jmczerk:
> So this is for the case when we can still run some GHA, but some are stuck? No objections, but I'm not especially informed either :)

The specific issue I had yesterday was that the queued actions timed out and failed, and there was no way that I knew of to manually kick them (once GitHub was healed) without a defined `workflow_dispatch` (which allows you to run the action through the GitHub GUI or with the `gh` cli).  So the new code was available, but I couldn't consume it downstream as the artifacts weren't published to Artifactory.

Though it sounds like based on what @yonghaoy is saying, I could have just created a release (once GitHub was healed) to get the new version bumped and published?  What is the process for creating a release?

wchamber:
Mechanically, it's through repo homepage, tags button, release button, draft a new release button, e.g. https://github.com/DataBiosphere/terra-common-lib/releases/new
Process-wise, I'm not sure. It looks like we have some automated github bot that creates new releases on every push, so doing a one-off manual release seems fine to me.

yonghaoy:
> > So this is for the case when we can still run some GHA, but some are stuck? No objections, but I'm not especially informed either :)
> 
> The specific issue I had yesterday was that the queued actions timed out and failed, and there was no way that I knew of to manually kick them (once GitHub was healed) without a defined `workflow_dispatch` (which allows you to run the action through the GitHub GUI or with the `gh` cli). So the new code was available, but I couldn't consume it downstream as the artifacts weren't published to Artifactory.
> 
> Though it sounds like based on what @yonghaoy is saying, I could have just created a release (once GitHub was healed) to get the new version bumped and published? What is the process for creating a release?

See: https://github.com/DataBiosphere/terra-common-lib#publishing

Github release is for  `lib-snapshot-release`, and I think we never use that. 
Auto release is for  `lib-snapshot-local` that is the one all services are using. I think at some point, once we feel confident to "officially" release this common-lib, we probably will start using the `-release` repo.
I think it might be helpful to enable workflow dispatch for `lib-snapshot-local`, not don't need it for `lib-snapshot-release`?

jmczerk:
> > > So this is for the case when we can still run some GHA, but some are stuck? No objections, but I'm not especially informed either :)
> > 
> > 
> > The specific issue I had yesterday was that the queued actions timed out and failed, and there was no way that I knew of to manually kick them (once GitHub was healed) without a defined `workflow_dispatch` (which allows you to run the action through the GitHub GUI or with the `gh` cli). So the new code was available, but I couldn't consume it downstream as the artifacts weren't published to Artifactory.
> > Though it sounds like based on what @yonghaoy is saying, I could have just created a release (once GitHub was healed) to get the new version bumped and published? What is the process for creating a release?
> 
> See: https://github.com/DataBiosphere/terra-common-lib#publishing
> 
> Github release is for `lib-snapshot-release`, and I think we never use that.
> Auto release is for `lib-snapshot-local` that is the one all services are using. I think at some point, once we feel confident to "officially" release this common-lib, we probably will start using the `-release` repo.
> I think it might be helpful to enable workflow dispatch for `lib-snapshot-local`, not don't need it for `lib-snapshot-release`?

Ah, I see what you are saying....  I hit the wrong file here, should have changed `develop-push.yml`, not `publish-release.yml`.

yonghaoy (approved):

PR #57, opened 2021-05-24
[PF-615] Modify DB retry utils behavior when max retries is exceeded

zloery:
Currently, the DB retry utils will throw an InterruptedException if the maximum number of retries for an operation is exceeded. This does not match the meaning of `InterruptedException`, so this change makes the function throw the most recent database error instead.

With this change, any call to `executeAndRetry` with a `retrySleep` duration of 0 will never throw an InterruptedException. This is useful for the short serializable transactions that several services run, as they will no longer need to surface an unnecessary checked exception.

The only existing usage of this function [is in RBS](https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/service/pool/PoolService.java#L115), where the exception is immediately surfaced. I can update that usage after this is merged, though it isn't strictly required because the DB exceptions are unchecked.

### DIFF
--- a/src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
+++ b/src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
@@ -21,7 +21,7 @@ public final class DatabaseRetryUtils {
    *
    * @param execute database operation to execute
    * @param retrySleep fixed retry sleep interval
-   * @param maxNumAttempts maximum retries
+   * @param maxNumAttempts maximum times this operation will be run
    * @param <T> database operation class
    * @return database operation class
    * @throws InterruptedException on thread interruption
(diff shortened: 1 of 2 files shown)


yonghaoy (approved):

src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
```diff
@@ -24,24 +25,27 @@ private DatabaseRetryUtils() {}
    * @param <T> database operation class
    * @return database operation class
    * @throws InterruptedException on thread interruption
+   * @throws DataAccessException throws the last DB operation error if maxNumAttempts is exceeded.
    */
   public static <T> T executeAndRetry(
       DatabaseOperation<T> execute, Duration retrySleep, int maxNumAttempts)
```
jaycarlton, line 31: As written, I think `maxNumAttempts` includes the first attempt, but it's documented as max retries

jaycarlton (approved):
is the plan to expand usage of this function?

src/main/java/bio/terra/common/db/DatabaseRetryUtils.java
zloery: good catch, updated doc to reflect that this is # attempts, not # retries.

zloery:
Yes, the plan is to use this function in WSM to support DB retries there

PR #59, opened 2021-07-09
Transaction retries and annotations

dvoet:


Shakespeared (approved):
LGTM (disclaimer: not an expert in retry logic)

snwessel (approved):
Looks good to me!

src/main/java/bio/terra/common/retry/transaction/TransactionRetryConfig.java
```diff
@@ -0,0 +1,121 @@
...
+  private int fastRetryMaxAttempts;
+  private Duration fastRetryMinBackOffPeriod;
+  private Duration fastRetryMaxBackOffPeriod;
+
+  private List<Class<? extends Throwable>> slowRetryExceptions;
+  private int slowRetryMaxAttempts;
+  private Duration slowRetryInitialInterval;
+  private double slowRetryMultiplier;
+
+  private BinaryExceptionClassifier slowRetryExceptionClassifier;
```
ddietterich: Why do these need to be configurable?

src/main/java/bio/terra/common/retry/transaction/TransactionRetryConfig.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  /** BinaryExceptionClassifier created using fastRetryExceptions */
+  public BinaryExceptionClassifier getFastRetryExceptionClassifier() {
+    return fastRetryExceptionClassifier;
+  }
+
+  @Override
+  public void afterPropertiesSet() throws Exception {
+    Objects.requireNonNull(this.fastRetryExceptions, "fastRetryExceptions must be specified");
```
ddietterich: I would prefer providing defaults rather than making errors. I predict that everyone will be coalesce using the same values and we will just be copying the same list from helmfile to helmfile. And when we find a new one, we will have N places to go fix.

src/main/java/bio/terra/common/retry/transaction/TransactionRetryInterceptorFactory.java
```diff
@@ -0,0 +1,71 @@
...
+
+@Configuration
+@EnableConfigurationProperties(TransactionRetryConfig.class)
+public class TransactionRetryInterceptorFactory {
+  /**
+   * Creates an interceptor that can be used in {@link
+   * org.springframework.retry.annotation.Retryable}: <code>
+   * @Retryable(interceptor = "transactionRetryInterceptor")</code>. Be sure to use {@link
+   * org.springframework.retry.annotation.EnableRetry}.
+   */
```
ddietterich: What is the cardinality of this? Factory makes it sound like there is more than one. So is it done once for each `@ReadTransaction` or `@WriteTransaction` or is it done on each entry into the retry code?

src/main/java/bio/terra/common/retry/CompositeBackOffPolicy.java
```diff
@@ -0,0 +1,77 @@
...
+  private final LinkedHashMap<BinaryExceptionClassifier, BackOffPolicy> backOffPoliciesByClassifier;
+
+  public CompositeBackOffPolicy(
+      LinkedHashMap<BinaryExceptionClassifier, BackOffPolicy> backOffPoliciesByClassifier) {
+    this.backOffPoliciesByClassifier = backOffPoliciesByClassifier;
+  }
+
+  @Override
+  public BackOffContext start(RetryContext context) {
+    Map<BinaryExceptionClassifier, BackOffContext> backOffContextMap = new HashMap<>();
```
ddietterich: Why is this a map? You don't use the map functionality. It could be an array of Pair.

src/main/java/bio/terra/common/retry/CompositeBackOffPolicy.java
```diff
@@ -0,0 +1,77 @@
...
+  // LinkedHashMap to ensure consistent behavior when an exception matches more than 1 policy
+  private final LinkedHashMap<BinaryExceptionClassifier, BackOffPolicy> backOffPoliciesByClassifier;
+
+  public CompositeBackOffPolicy(
+      LinkedHashMap<BinaryExceptionClassifier, BackOffPolicy> backOffPoliciesByClassifier) {
+    this.backOffPoliciesByClassifier = backOffPoliciesByClassifier;
+  }
+
+  @Override
+  public BackOffContext start(RetryContext context) {
```
ddietterich, line 38: Suggest a comment like: this method gets called on entry to a `@StartTransaction` or `@EndTransaction` to set up the dynamic context for backing off on a retry. Since we have slow and fast retry, we need to setup context for both and package that inside a composite BackoffContext subclass.

src/main/java/bio/terra/common/retry/transaction/TransactionRetryInterceptorFactory.java
```diff
@@ -0,0 +1,71 @@
...
+  private BackOffPolicy createTransactionBackOffPolicy(TransactionRetryConfig config) {
+    UniformRandomBackOffPolicy fastBackOffPolicy = new UniformRandomBackOffPolicy();
+    fastBackOffPolicy.setMaxBackOffPeriod(config.getFastRetryMaxBackOffPeriod().toMillis());
+    fastBackOffPolicy.setMinBackOffPeriod(config.getFastRetryMinBackOffPeriod().toMillis());
+
+    ExponentialBackOffPolicy slowBackOffPolicy = new ExponentialBackOffPolicy();
+    slowBackOffPolicy.setInitialInterval(config.getSlowRetryInitialInterval().toMillis());
+    slowBackOffPolicy.setMultiplier(config.getSlowRetryMultiplier());
+
+    LinkedHashMap<BinaryExceptionClassifier, BackOffPolicy> backOffPolicies = new LinkedHashMap<>();
```
ddietterich: Why not make this a list?

src/main/java/bio/terra/common/retry/transaction/TransactionRetryConfig.java
```diff
@@ -0,0 +1,121 @@
...
+import org.springframework.classify.BinaryExceptionClassifier;
+
+/**
+ * Configuration settings for how database transactions are retried. There are 2 classes of retries,
+ * fast and slow. Fast retries usually measure in milliseconds with a constant retry period (with
+ * some random jitter) and many attempts. Slow retries usually measure in seconds with an
+ * exponential back off and few attempts.
+ */
+@ConfigurationProperties(prefix = "terra.common.retry.transaction")
+public class TransactionRetryConfig implements InitializingBean {
```
ddietterich: There is a pattern of separating properties from configuration that is used in TCL.
I'm not sure that should apply here. @gjuggler pioneered that split, I think, so maybe he can comment or provide a pointer.

I think if you provided defaults for the exceptions, you would not need to derive from InitializingBeam.

src/main/java/bio/terra/common/retry/transaction/TransactionRetryInterceptorFactory.java
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Creates an interceptor that can be used in {@link
+   * org.springframework.retry.annotation.Retryable}: <code>
+   * @Retryable(interceptor = "transactionRetryInterceptor")</code>. Be sure to use {@link
+   * org.springframework.retry.annotation.EnableRetry}.
+   */
+  @Bean("transactionRetryInterceptor")
+  public MethodInterceptor getTransactionRetryInterceptor(TransactionRetryConfig config) {
+    return RetryInterceptorBuilder.stateless()
+        .retryPolicy(createTransactionRetryPolicy(config))
```
ddietterich: Why we need to create new retry and backoff policies each time?
IIUC those are read-only once created. All read-write is done in the per-thread retry context.

ddietterich (changes requested):
Mostly questions. I'd like more documentation.

I read some Spring doc'n to understand what this was doing. But the Spring docs are not very good. And the interceptor technique (nice, btw) wasn't mentioned in what I read.

src/main/java/bio/terra/common/retry/transaction/TransactionRetryConfig.java
dvoet: because I feel some uncertainty about what exceptions should be retried and I am not certain the same exceptions should be retried in all projects

src/main/java/bio/terra/common/retry/transaction/TransactionRetryConfig.java
dvoet: at the end of the day, there are still N places, N+1 actually. It needs to be fixed in the library and then the latest version of the library needs to be rolled out to N projects and compatibility problems fixed.

src/main/java/bio/terra/common/retry/transaction/TransactionRetryInterceptorFactory.java
dvoet: I think it is just once. What do you call a class that has a method with a @Bean annotation?

src/main/java/bio/terra/common/retry/CompositeBackOffPolicy.java
dvoet: correct

dvoet:
@ddietterich back to you, I think I addressed most of your comments. I tried to remove the linked hash map but ended up needing the map functionality

src/main/java/bio/terra/common/retry/CompositeBackOffPolicy.java
```diff
@@ -1,9 +1,8 @@
 package bio.terra.common.retry;
 
 import bio.terra.common.db.DatabaseRetryUtils;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.Map;
+import java.util.*;
```
ddietterich: nit: we don't use `*` here. There is an IntelliJ setting to force it not to generate them: `Preferences > Editor > Code Style > Java` and then set very large numbers for the class count and name count.

ddietterich (approved):
LGTM, thank you!

One code format nit

PR #60, opened 2021-09-10
[PF-976] Use refactored Stairway

ddietterich:
The refactor makes `Stairway` an interface, so the `StairwayBuilder` moves into its own class.
Also, the pubsub code is factored out of the stairway library and into stairway-gcp library.

I moved dynamic construction of the pubsub queues into TCL. The target state is for pubsub objects to be created by Terraform and passed in rather than do dynamic creation. That is tracked as [PF-558](https://broadworkbench.atlassian.net/browse/PF-558)

I tested this by integrating with WSM and running the WSM full regression tests.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,6 @@ plugins {
 }
 
 repositories {
-    mavenLocal()
     mavenCentral()
     maven {
         url 'https://broadinstitute.jfrog.io/broadinstitute/libs-release/'
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,7 @@ plugins {
 }
 
 repositories {
+    mavenLocal()
     mavenCentral()
     maven {
         url 'https://broadinstitute.jfrog.io/broadinstitute/libs-release/'
@@ -41,14 +42,17 @@ dependencies {
     }
     implementation platform('com.google.cloud:libraries-bom:20.2.0')
 
-    implementation group: 'bio.terra', name: 'stairway', version: '0.0.56-SNAPSHOT'
-    implementation group: 'org.postgresql', name: 'postgresql', version: '42.2.18'
+    // Publish the stairway library for common lib users to minimize rollout overhead
+    // of new Stairways.
+    api group: 'bio.terra', name: 'stairway-gcp', version: '0.0.66-SNAPSHOT'
 
+    implementation group: 'org.postgresql', name: 'postgresql', version: '42.2.18'
     implementation group: 'ch.qos.logback.contrib', name: 'logback-json-classic', version: '0.1.5'
     implementation group: 'ch.qos.logback.contrib', name: 'logback-jackson', version: '0.1.5'
     implementation group: 'com.google.cloud', name: 'google-cloud-core'
     implementation group: 'com.google.code.findbugs', name: 'annotations', version: '3.0.1'
     implementation group: 'com.google.apis', name: 'google-api-services-logging', version: 'v2-rev20210507-1.31.0'
+    implementation group: 'com.google.cloud', name: 'google-cloud-pubsub'
     implementation group: 'io.kubernetes', name: 'client-java', version: '10.0.0'
     implementation group: 'io.opencensus', name: 'opencensus-exporter-trace-stackdriver', version: opencensus
     // commons-dbcp2 and -lang3 versions are managed by the Spring Boot plugin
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -17,6 +17,7 @@ plugins {
 }
 
 repositories {
+    mavenLocal()
```
jaycarlton: I think you want to remove this before merging.

src/main/java/bio/terra/common/stairway/StairwayProperties.java
```diff
@@ -14,44 +14,57 @@
...
-    return maxParallelFlights;
-  }
+  /** PubSub topic to use for stairway work queue. It must exist in the current GCP project */
+  private String gcpPubSubTopicId;
 
-  public Duration getQuietDownTimeout() {
-    return quietDownTimeout;
-  }
+  /**
+   * PubSub subscription to use for stairway work queue. It must exist in the current GCP project
```
wchamber: can we document that if these are null, we use the clusterNameSuffix to create topics/subscriptions?
Consider also renaming clusterNameSuffix to make it clear that it constrols creating a pubsub queue.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -151,6 +187,10 @@ public Stairway get() {
   }
 
   private String getClusterName() {
+    if (stairwayProperties.getClusterNameSuffix() == null) {
+      throw new IllegalArgumentException(
+          "Cluster name suffix property is required for Stairway work queue creation");
```
wchamber: nit: I think this assertion might be clearer if it were inlined in setupWorkQueue. As-is, I wouldn't really expect this validation to happen in a method named `getClusterName`, and I initially thought the comment on setUpWorkQueue was incorrect.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -63,24 +68,55 @@ public void initialize(DataSource dataSource, Object context, List<StairwayHook>
...
+   * Set up the Stairway work queue. There are two ways we get a work queue. If both the gcp topicId
+   * and subscriptionId are passed as parameters, then we use those. Otherwise, if the
+   * clusterNameSuffix is present, then we create the queue. It is an error to try to run in
+   * Kubernetes without a work queue.
+   */
+  private GcpPubSubQueue setupWorkQueue() {
+    try {
+      String topicId = stairwayProperties.getGcpPubSubTopicId();
+      String subscriptionId = stairwayProperties.getGcpPubSubSubscriptionId();
+      if (topicId == null || subscriptionId == null) {
```
wchamber: optional: if both aren't null, consider asserting that clusterNamePrefix is null to prevent cases where it's not obvious what should happen.

wchamber (approved):

src/main/java/bio/terra/common/stairway/StairwayProperties.java
```diff
@@ -87,4 +100,22 @@ public Duration getCompletedFlightRetention() {
...
     this.completedFlightRetention = completedFlightRetention;
   }
+
+  public String getGcpPubSubTopicId() {
+    return gcpPubSubTopicId;
+  }
+
+  public StairwayProperties gcpPubSubTopicId(String gcpPubSubTopicId) {
+    this.gcpPubSubTopicId = gcpPubSubTopicId;
+    return this;
```
jaycarlton: Do you need these fluent-style methods for all the properties?

jaycarlton (approved):

src/main/java/bio/terra/common/stairway/StairwayProperties.java
ddietterich: I tried to improve the documentation in the properties file.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
ddietterich: I rearranged the logic to test for the two allowed cases and throw with an explanatory error.
I consolidated getClusterName into setupWorkQueue.

src/main/java/bio/terra/common/stairway/StairwayProperties.java
ddietterich: No. I fixed them to match the other accessors.

build.gradle
ddietterich: Thank you!

PR #61, opened 2021-09-23
[PF-881] Expand set of error codes which are retried in SamRetry

zloery:
Currently, SamRetry will only retry 500 errors on calls to Sam. There are a few other codes we should retry as well, as well as timeouts which will fail with error code 0. I believe other connection issues will also fail with code 0, but have yet to see any across the services using client libraries to make calls.


src/main/java/bio/terra/common/sam/SamRetry.java
```diff
@@ -27,11 +26,13 @@
   private static final Duration OPERATION_TIMEOUT = Duration.ofSeconds(300);
   private final Instant operationTimeout;
 
+  // Sam calls which timeout will throw ApiExceptions wrapping SocketTimeoutExceptions and will have
+  // an errorCode 0. This isn't a real HTTP status code, but we can check for it anyway.
+  private static final int TIMEOUT_STATUS_CODE = 0;
```
wchamber, line 32: Is there anything else in the ApiException we can use to check if it's a time out exception? It might be better to be more explicit rather than retry anything that doesn't have a status code and defaults to 0.
Not sure how you can figure this out other than wait for Sam to socket timeout and look at the exception...

src/main/java/bio/terra/common/sam/SamRetry.java
```diff
@@ -90,7 +91,11 @@ public static void retry(SamVoidFunction function, Duration timeout)
   }
 
   private boolean isRetryable(ApiException apiException) {
-    return (apiException.getCode() == HttpStatus.INTERNAL_SERVER_ERROR.value());
+    return apiException.getCode() == TIMEOUT_STATUS_CODE
+        || apiException.getCode() == HttpStatus.SC_INTERNAL_SERVER_ERROR
+        || apiException.getCode() == HttpStatus.SC_BAD_GATEWAY
+        || apiException.getCode() == HttpStatus.SC_SERVICE_UNAVAILABLE
```
wchamber, line 109: This one's a little dangerous because if Sam is truly overloaded, we're just going to retry hammer it, making the overload worse. Client side throttling would be a better solution, but there's nothing off-the-shelf to use here.
Consider not retrying this case, very long retry times available in a retry mode that can tolerate long delays (i.e. during a flight), or just being really mean to Sam and retrying anyway.

src/main/java/bio/terra/common/sam/exception/SamConnectionException.java
```diff
@@ -0,0 +1,16 @@
+package bio.terra.common.sam.exception;
+
+import bio.terra.common.exception.BadRequestException;
+
+/**
+ * This exception is thrown when a request fails due to a connection issue, most often a timeout.
+ */
+public class SamConnectionException extends BadRequestException {
```
wchamber: that doesn't seem like it's a bad request? Probably a service unavailable extension?

src/main/java/bio/terra/common/sam/exception/SamExceptionFactory.java
```diff
@@ -18,6 +18,9 @@
 public class SamExceptionFactory {
   private static final ObjectMapper objectMapper = new ObjectMapper();
   private static Logger logger = LoggerFactory.getLogger(SamExceptionFactory.class);
+  // Requests which time out will have status code 0. This is not a real HTTP status code, but is
+  // still a useful signal.
+  private static final int TIMEOUT_ERROR_CODE = 0;
```
wchamber: let's at least de-dupe this inference of 0 -> timeout to a shared function.

src/main/java/bio/terra/common/sam/SamRetry.java
zloery: Good point, we can also look at the `cause` field which will have a `SocketTimeoutException` for timeouts. 
(and yep, that was how I found the 0 status code)

src/main/java/bio/terra/common/sam/SamRetry.java
zloery: My inclination is to just be mean here, in practice I generally see this returned when hitting a pod which is still starting up, which is definitely a retryable case. I don't whether Sam also throws this under high load though, when retrying would be harmful.

src/main/java/bio/terra/common/sam/exception/SamConnectionException.java
zloery: Whoops, that's what I get for copying other exceptions. Changed to generic internal server exception

src/main/java/bio/terra/common/sam/exception/SamExceptionFactory.java
```diff
@@ -1,6 +1,7 @@
 package bio.terra.common.sam.exception;
 
 import bio.terra.common.exception.ErrorReportException;
+import bio.terra.common.sam.SamRetry;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.api.client.http.HttpStatusCodes;
```
marikomedlock: nit: Would be good to standardize on one http status code enum.

marikomedlock (commented):
Not necessarily this PR, but if you are going to retry wrapped socket exceptions, I think it makes sense to also retry unwrapped socket exceptions.

- In the CLI, I've seen SAM socket timeouts directly, not wrapped in a SAM ApiException.
- I've also seen SAM socket timeouts wrapped in a WSM ApiException, which I think means that WSM got an un-wrapped SAM socket timeout.
- I don't remember seeing a SAM socket timeout wrapped in a SAM ApiException, but I think that means SAM got a socket timeout somewhere internally (e.g. talking to its DB?). Though since you've seen this, I'll plan to make the CLI retry logic consistent.

marikomedlock (approved):
Summary of offline conversation with Zach:

- SAM client method signatures don't include an IOException or SocketTimeoutException, so they won't throw those checked exceptions. So no need to retry these in common lib either.
- CLI is retrying these, but that may be leftover from some code that makes HTTP requests to SAM directly without the client library. I'm going to take a ticket to reevaluate that logic in the CLI and bring it in line with what we've got here.

src/main/java/bio/terra/common/sam/SamRetry.java
wchamber: SG, let's leave a comment about that choice.

src/main/java/bio/terra/common/sam/SamRetry.java
```diff
@@ -53,6 +54,11 @@ protected SamRetry(Duration timeout) {
     R apply() throws ApiException, InterruptedException;
   }
 
+  public static boolean isTimeoutException(ApiException apiException) {
+    return apiException.getCode() == TIMEOUT_STATUS_CODE
+        && apiException.getCause() instanceof SocketTimeoutException;
```
wchamber, line 65: worth a quick comment and on what this means to have a timeout exception and why we think it happens today.

src/main/java/bio/terra/common/sam/exception/SamTimeoutException.java
```diff
@@ -0,0 +1,17 @@
+package bio.terra.common.sam.exception;
+
+import bio.terra.common.exception.InternalServerErrorException;
+
+/**
+ * This exception is thrown when a request fails due to a timeout. Timeouts are generally retried,
+ * but this may be thrown if repeated retries fail.
+ */
+public class SamTimeoutException extends InternalServerErrorException {
```
wchamber: I think this a 503 instead of a 500 case, but I'm not an http code expert, so use your judgment.

wchamber (approved):

PR #62, opened 2021-12-03
[DDO-1778] Add metrics exporting support

jack-r-warren:
Allows Spring apps using this library to component scan `bio.terra.common.metrics` to export any OpenCensus metrics in Prometheus format on a different port.

- I went with `metrics` over something like `prometheus` to fit with the `tracing` naming convention, but happy to change that
- I did some research on Prometheus port number standards to come up with a good default is within the standard Prometheus range but won't clobber anything

I tested this with the usual mavenLocal process for Buffer, just by scanning the package all its metrics got exported properly to `localhost:9098/metrics`

### DIFF
--- a/src/main/java/bio/terra/common/prometheus/PrometheusComponent.java
+++ b/src/main/java/bio/terra/common/prometheus/PrometheusComponent.java
@@ -1,4 +1,4 @@
-package bio.terra.common.metrics;
+package bio.terra.common.prometheus;
 
 import io.opencensus.exporter.stats.prometheus.PrometheusStatsCollector;
 import io.prometheus.client.exporter.HTTPServer;
@@ -11,24 +11,24 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
-/** A Spring Component that exposes any OpenCensus metrics via a {@link HTTPServer}. */
+/** A Spring Component that exposes any OpenCensus metrics via a Prometheus {@link HTTPServer}. */
 @Component
-public class MetricsComponent {
-  private final Logger logger = LoggerFactory.getLogger(MetricsComponent.class);
-  private final MetricsProperties metricsProperties;
+public class PrometheusComponent {
+  private final Logger logger = LoggerFactory.getLogger(PrometheusComponent.class);
+  private final PrometheusProperties prometheusProperties;
 
   private HTTPServer prometheusServer;
 
   @Autowired
-  public MetricsComponent(MetricsProperties metricsProperties) {
-    this.metricsProperties = metricsProperties;
+  public PrometheusComponent(PrometheusProperties prometheusProperties) {
+    this.prometheusProperties = prometheusProperties;
   }
 
   @PostConstruct
   private void startEndpointServer() {
     logger.info(
-        "Prometheus metrics endpoint enabled: {}", metricsProperties.isPrometheusEndpointEnabled());
-    if (!metricsProperties.isPrometheusEndpointEnabled()) {
+        "Prometheus metrics endpoint enabled: {}", prometheusProperties.isEndpointEnabled());
+    if (!prometheusProperties.isEndpointEnabled()) {
       return;
     }
     try {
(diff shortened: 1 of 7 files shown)


yonghaoy (commented):
metrics as the package is bit confusing/general. 
Seems this is just a library for `prometheus`. Shall we rename this to `prometheus`?

src/main/java/bio/terra/common/prometheus/PrometheusComponent.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.annotation.PreDestroy;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/** A Spring Component that exposes any OpenCensus metrics via a Prometheus {@link HTTPServer}. */
+@Component
+public class PrometheusComponent {
+  private final Logger logger = LoggerFactory.getLogger(PrometheusComponent.class);
```
yonghaoy: nit: Logger can be `static`

yonghaoy (approved):

jmczerk (approved):
So with this in place, I assume that the existing Resource Buffer PR becomes a matter of just configuring this?

yonghaoy:
> So with this in place, I assume that the existing Resource Buffer PR becomes a matter of just configuring this?

Next step after pr merge: Bump library version in Buffer, and also add that component scan for the new package https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/java/bio/terra/buffer/app/Main.java#L12

@jmczerk btw, for Verily deployment, if you want to turn this off, you can set something like: `TERRA_COMMON_PROMETHEUS_ENDPOINT_ENABLED=false `in env var.

jack-r-warren:
Sounds good, seems like the same thing I did for testing. Also re deployment, note that I'm only enabling scraping of the prometheus endpoint for the "live" environments, not the "personal" ones. So the metrics endpoint will run like normal by default but our Prometheus config won't start looking at it.

yonghaoy:
> Sounds good, seems like the same thing I did for testing. Also re deployment, note that I'm only enabling scraping of the prometheus endpoint for the "live" environments, not the "personal" ones. So the metrics endpoint will run like normal by default but our Prometheus config won't start looking at it.

SG!  Also, the default value is `enabled=true` https://github.com/DataBiosphere/terra-common-lib/pull/62/files#diff-b0a0907e8593139f68e57e750575bf481da2b2e7a43f9cf3cccd72566a43a986R15
Then if you want to enable that in personal, GHA integration test, unite test, we would need to change to override those values. 
Not sure if that is what you intended or you want to set the default value to `false`, then just change. the live.yaml in helmfile.

jack-r-warren:
> > Sounds good, seems like the same thing I did for testing. Also re deployment, note that I'm only enabling scraping of the prometheus endpoint for the "live" environments, not the "personal" ones. So the metrics endpoint will run like normal by default but our Prometheus config won't start looking at it.
> 
> SG! Also, the default value is `enabled=true` [#62 (files)](https://github.com/DataBiosphere/terra-common-lib/pull/62/files#diff-b0a0907e8593139f68e57e750575bf481da2b2e7a43f9cf3cccd72566a43a986R15) Then if you want to enable that in personal, GHA integration test, unite test, we would need to change to override those values. Not sure if that is what you intended or you want to set the default value to `false`, then just change. the live.yaml in helmfile.

Yeah, there's effectively two things to toggle: whether Buffer hosts the endpoint (it does by default everywhere, exactly what you linked) and whether Prometheus knows to look at the endpoint (that's controlled via Helm exactly like you suggested (https://github.com/broadinstitute/terra-helmfile/blob/DDO-1729-buffer-metrics/values/app/buffer/live.yaml#L15)). My thinking was "the app should operate the same everywhere so there's no surprises, but we only want to actually collect the metrics from live environments"

yonghaoy:
> > > Sounds good, seems like the same thing I did for testing. Also re deployment, note that I'm only enabling scraping of the prometheus endpoint for the "live" environments, not the "personal" ones. So the metrics endpoint will run like normal by default but our Prometheus config won't start looking at it.
> > 
> > 
> > SG! Also, the default value is `enabled=true` [#62 (files)](https://github.com/DataBiosphere/terra-common-lib/pull/62/files#diff-b0a0907e8593139f68e57e750575bf481da2b2e7a43f9cf3cccd72566a43a986R15) Then if you want to enable that in personal, GHA integration test, unite test, we would need to change to override those values. Not sure if that is what you intended or you want to set the default value to `false`, then just change. the live.yaml in helmfile.
> 
> Yeah, there's effectively two things to toggle: whether Buffer hosts the endpoint (it does by default everywhere, exactly what you linked) and whether Prometheus knows to look at the endpoint (that's controlled via Helm exactly like you suggested (https://github.com/broadinstitute/terra-helmfile/blob/DDO-1729-buffer-metrics/values/app/buffer/live.yaml#L15)). My thinking was "the app should operate the same everywhere so there's no surprises, but we only want to actually collect the metrics from live environments"

My question was: do we want the default value to be 1: false, and turn it off in personal/integration/local servers. 
OR 2: to be true, and just turn if off in Broad's live deployment. 
And your current approche is 1, because if we leave that value unset, it will use the Java value which is the code I referred above.

PR #65, opened 2022-03-25
DR-2496 Remove message HTML escaping from the base ErrorReportException class

nmalfroy:
I also removed the unit test since now there's not much point to it


zloery (approved):
nit: the `HtmlEscapers` import is now unused, remove it

You can also run `gradlew spotlessApply` to run the Spotless linter, which should auto-fix that and any other formatting issues.

nmalfroy:
shoot, you're right. I thought I had but clearly not.  just updated

jmczerk (approved):
LGTM

PR #66, opened 2022-04-11
[PF-1531] Update dependencies, add lockfile

zloery:
This bumps the version of several dependencies with identified CVEs, as we rely on these in downstream services. The main changes are:

- io.kubernetes:client-java 10.0.0 -> 15.0.0
- org.postgresql:postgresql 42.2.12 -> 42.3.3

I also bumped the Spring Boot packages used for the test application 2.4.0 -> 2.6.6. I don't think it's strictly necessary since this isn't a deployed service, but we might as well keep this up to date.

This also enforces using a gradle lockfile to track dependency versions. There's already a [section in the README](https://github.com/DataBiosphere/terra-common-lib/blob/develop/README.md#dependencies) about doing this, but it looks like we didn't actually create the file.


yuhuyoyo (approved):

zloery:
Should be fixed now ~~Actually looks like this broke a few things - feel free to hold off reviewing until I get this fixed~~

build.gradle
```diff
@@ -66,8 +66,8 @@ dependencies {
     implementation group: 'org.springframework' , name: 'spring-context', version: '5.3.7'
     implementation group: 'org.springframework' , name: 'spring-core', version: '5.3.7'
     implementation group: 'org.springframework.retry', name: 'spring-retry', version: '1.3.1'
-    implementation group: 'org.springframework.boot', name: 'spring-boot-starter-data-jdbc', version: '2.4.5'
-    implementation group: 'org.springframework.boot', name: 'spring-boot-starter-web', version: '2.4.5'
+    implementation group: 'org.springframework.boot', name: 'spring-boot-starter-data-jdbc', version: '2.6.6'
```
jaycarlton, line 69: For the other repos, it seems like we use a platform for the Spring dependencies. I wonder why they're explicit here. Is it safe to pull in TCL code referring to a different version of Spring than the service's?

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
```diff
@@ -63,8 +62,6 @@ class GoogleJsonLayout extends JsonLayoutBase<ILoggingEvent> {
   // A reference to the current Spring app context, on order to pull out the spring.application.name
   // and spring.application.version variable for inclusion in JSON output.
   private ConfigurableApplicationContext applicationContext;
-  // A Gson instance to support converting Gson-type payloads into Jackson nodes.
```
jaycarlton, line 66: So we're still using Gson, just not this instance?

jaycarlton (approved):

build.gradle
zloery: I think it's because we're only pulling in a few packages here since TCL is just a library (though it has a test app), rather than the full set to deploy a service.

I'm not actually sure how safe updating these packages out from under a service will be. My plan was to pull in the updated TCL version and upgrade Spring Boot to `2.6.6` at the same time for all the downstream services, so I didn't think about it very hard.

src/main/java/bio/terra/common/logging/GoogleJsonLayout.java
zloery: Yep. This still supports Gson (and `toJsonMap` still handles `gson.JsonObject`s), this is just cleaning up an unused field.

PR #71, opened 2022-04-19
[No ticket] Delete unused action

zloery:
I noticed this action is only triggered on releases, but we never make releases through Github for this library. It looks like it's only ever been run once: https://github.com/DataBiosphere/terra-common-lib/actions/workflows/publish-release.yml


jmczerk (approved):

jaycarlton (approved):
How do we do releases for this lib? Why wouldn't we want to do it the same way as CRL?

zloery:
We still bump the version and publish artifacts to Artifactory, but it's triggered on PRs merging into the main branch via [a different action](https://github.com/DataBiosphere/terra-common-lib/blob/develop/.github/workflows/develop-push.yml) instead of on a Github release being created. That's the same pattern we currently use in CRL - we used to publish Github releases there, but [don't anymore](https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/122)

PR #73, opened 2022-05-31
new auth token handling

dvoet:
This PR updates authenticated requests. I feel it is important to deprecate `AuthenticatedUserRequest` because it is misleading. It extracted the email address and user id from the access token which are not alway correct. 

It is important to separate the use cases that only need an access token from those that actually need user information from sam. Querying sam for every request is a waste of electricity and is bad for the environment as the majority of api calls don't need that detail. 

[TokenAuthenticatedRequest example](https://github.com/DataBiosphere/terra-drs-hub/blob/ac258fc4525ae8be2443d924c9e2ea4f1af46792/service/src/main/java/bio/terra/drshub/controllers/DrsHubApiController.java#L36)

[SamUserAuthenticatedRequest example](https://github.com/DataBiosphere/terra-external-credentials-manager/blob/SamUserAuthenticatedRequest/service/src/main/java/bio/terra/externalcreds/controllers/OidcApiController.java#L51)


src/main/java/bio/terra/common/iam/SamAuthenticatedUserRequestFactory.java
```diff
@@ -0,0 +1,74 @@
...
+import org.broadinstitute.dsde.workbench.client.sam.ApiClient;
+import org.broadinstitute.dsde.workbench.client.sam.ApiException;
+import org.broadinstitute.dsde.workbench.client.sam.api.UsersApi;
+import org.broadinstitute.dsde.workbench.client.sam.model.UserStatusInfo;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Component;
+
+/**
+ * Used verify a user is enabled and to get accurate email and id for the user from Sam. Email and
```
pshapiro4broad: typo?
```suggestion
 * Use to verify that a user is enabled and to get accurate email and id for the user from Sam. Email and
```

src/main/java/bio/terra/common/iam/SamAuthenticatedUserRequestFactory.java
```diff
@@ -0,0 +1,74 @@
...
+ * id embedded in the request's authentication token are not always accurate (in the case of pet
+ * identities) and should not be used. Please do not use by default on every request, instead use
+ * {@link TokenAuthenticatedRequestFactory} Use this factory only when user id and email are
```
pshapiro4broad:
```suggestion
 * {@link TokenAuthenticatedRequestFactory}. Use this factory only when user id and email are
```

src/main/java/bio/terra/common/iam/SamAuthenticatedUserRequestFactory.java
```diff
@@ -0,0 +1,74 @@
...
+    this.tokenAuthenticatedRequestFactory = tokenAuthenticatedRequestFactory;
+  }
+
+  public SamUserAuthenticatedRequest from(
+      HttpServletRequest servletRequest, ApiClient samApiClient) {
+    return from(tokenAuthenticatedRequestFactory.from(servletRequest), samApiClient);
+  }
+
+  public SamUserAuthenticatedRequest from(
+      TokenAuthenticatedRequest tokenAuthenticatedRequest, ApiClient samApiClient) {
```
pshapiro4broad: It's a minor thing but the convention I usually see with overloaded methods is to have the identical parameters first.
```suggestion
  public SamUserAuthenticatedRequest from(
      ApiClient samApiClient, TokenAuthenticatedRequest tokenAuthenticatedRequest) {
```

src/main/java/bio/terra/common/iam/SamAuthenticatedUserRequestFactory.java
```diff
@@ -0,0 +1,74 @@
...
+    SamUserAuthenticatedRequest.Builder builder =
+        SamUserAuthenticatedRequest.builder().setTokenRequest(tokenAuthenticatedRequest);
+
+    samApiClient.setAccessToken(tokenAuthenticatedRequest.getToken());
```
pshapiro4broad: It seems a bit odd to modify the client the caller passes in. Do we do this sort of thing elsewhere in tcl?

We could have the caller pass in the sam base path instead of `ApiClient`. Or extract it from `samApiClient` and make a new api client using it.

src/main/java/bio/terra/common/iam/SamAuthenticatedUserRequestFactory.java
```diff
@@ -0,0 +1,74 @@
...
+    try {
+      UserStatusInfo userStatusInfo = usersApi.getUserStatusInfo();
+      if (!userStatusInfo.getEnabled()) {
+        throw new UnauthorizedException("User is disabled, please contact Terra support");
+      }
+      return builder
+          .setEmail(userStatusInfo.getUserEmail())
+          .setSubjectId(userStatusInfo.getUserSubjectId())
+          .build();
+    } catch (final IllegalStateException e) {
```
pshapiro4broad: I'm trying to figure out where this exception might be thrown from. It looks like `build()` is already set up to throw `UnauthorizedException()` if a field is missing. Do you have a test case that shows that this exception can be thrown?

src/main/java/bio/terra/common/iam/SamAuthenticatedUserRequestFactory.java
```diff
@@ -0,0 +1,74 @@
...
+    SamUserAuthenticatedRequest.Builder builder =
+        SamUserAuthenticatedRequest.builder().setTokenRequest(tokenAuthenticatedRequest);
+
+    samApiClient.setAccessToken(tokenAuthenticatedRequest.getToken());
+    UsersApi usersApi = getUsersApi(samApiClient);
+
+    try {
+      UserStatusInfo userStatusInfo = usersApi.getUserStatusInfo();
+      if (!userStatusInfo.getEnabled()) {
+        throw new UnauthorizedException("User is disabled, please contact Terra support");
```
pshapiro4broad: Is there a reason to not use `SamUnauthorizedException` here (and elsewhere in this PR)? (I don't know the reasons for `SamUnauthorizedException` vs `UnauthorizedException` but it seems like we should use it if we already have it, and are using it elsewhere.)

src/main/java/bio/terra/common/iam/SamUserAuthenticatedRequest.java
```diff
@@ -0,0 +1,120 @@
...
+
+import bio.terra.common.exception.UnauthorizedException;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import java.util.Objects;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+/** Class representing the identity of an authenticated user. */
+@JsonDeserialize(builder = SamUserAuthenticatedRequest.Builder.class)
+public class SamUserAuthenticatedRequest {
```
pshapiro4broad: Something I've always wondered about `AuthenticatedUserRequest`, is why is `request` in its name? It doesn't seem like a request. In this case, too, I wonder about the ordering of `Sam` `User` and `Authenticated`. I would expect `SamAuthenticatedUser` (or `SamUser`, since it seem redundant to say we have a sam user that's also authenticated, since we couldn't have a sam user if it wasn't also authenticated)

On the other hand, if we only use Sam for authentication, using `AuthenticatedUser` would also work.

src/main/java/bio/terra/common/iam/SamUserAuthenticatedRequest.java
```diff
@@ -0,0 +1,120 @@
...
+        && Objects.equals(getTokenRequest(), that.getTokenRequest());
+  }
+
+  @Override
+  public int hashCode() {
+    return new HashCodeBuilder()
+        .append(getEmail())
+        .append(getSubjectId())
+        .append(getTokenRequest())
+        .toHashCode();
```
pshapiro4broad: Is there a reason to use `HashCodeBuilder` instead of the Java builtin API for this? Also, why use getters here?
```suggestion
    return Objects.hash(email, subjectId, tokenRequest);
```

src/main/java/bio/terra/common/iam/TokenAuthenticatedRequest.java
```diff
@@ -0,0 +1,72 @@
...
+
+import bio.terra.common.exception.UnauthorizedException;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import java.util.Objects;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+/** Class representing the bearer token of the request. */
+@JsonDeserialize(builder = TokenAuthenticatedRequest.Builder.class)
+public class TokenAuthenticatedRequest {
```
pshapiro4broad: There doesn't seem to be anything request specific about this code, or authorization header specific. It looks like it's a POJO that stores a token. If so, why not call it `BearerToken`?

src/main/java/bio/terra/common/iam/TokenAuthenticatedRequest.java
```diff
@@ -0,0 +1,72 @@
...
+      this.token = token;
+      return this;
+    }
+
+    /**
+     * Build method returns a constructed, immutable {@link TokenAuthenticatedRequest} instance.
+     *
+     * @throws UnauthorizedException if any field are not properly initialized
+     */
+    public TokenAuthenticatedRequest build() {
```
pshapiro4broad: What is the benefit of having a builder for a class with a single field? Is it better in some way to throw in `build()` vs throwing in the constructor?

src/main/java/bio/terra/common/iam/TokenAuthenticatedRequestFactory.java
```diff
@@ -0,0 +1,28 @@
...
+ * token in the Authorization header.
+ *
+ * <p>This factory is all that should be required to propagate the caller's identity in most cases.
+ * In cases where Sam is not otherwise called to check the caller's access or the caller's email or
+ * id are required, seel{@link SamAuthenticatedUserRequestFactory}.
+ */
+@Component
+public class TokenAuthenticatedRequestFactory {
+
+  static final String AUTHORIZATION = "Authorization";
```
pshapiro4broad: It looks like this is only visible for testing purposes
```suggestion
  @VisisbleForTesting
  static final String AUTHORIZATION = "Authorization";
```

src/main/java/bio/terra/common/iam/SamAuthenticatedUserRequestFactory.java
```diff
@@ -0,0 +1,74 @@
...
+
+  public SamUserAuthenticatedRequest from(
+      HttpServletRequest servletRequest, ApiClient samApiClient) {
+    return from(tokenAuthenticatedRequestFactory.from(servletRequest), samApiClient);
+  }
+
+  public SamUserAuthenticatedRequest from(
+      TokenAuthenticatedRequest tokenAuthenticatedRequest, ApiClient samApiClient) {
+    SamUserAuthenticatedRequest.Builder builder =
+        SamUserAuthenticatedRequest.builder().setTokenRequest(tokenAuthenticatedRequest);
```
pshapiro4broad: This is a case where `var` works well
```suggestion
    var builder = SamUserAuthenticatedRequest.builder().setTokenRequest(tokenAuthenticatedRequest);
```

src/test/java/bio/terra/common/iam/SamUserAuthenticatedRequestFactoryTest.java
```diff
@@ -0,0 +1,74 @@
...
+  @Test
+  public void notFoundUser() throws ApiException {
+    TestFactory factory = new TestFactory();
+    when(factory.usersApi.getUserStatusInfo())
+        .thenThrow(new ApiException(HttpStatus.NOT_FOUND.value(), "not found"));
+
+    assertThrows(UnauthorizedException.class, () -> factory.from(TOKEN, new ApiClient()));
+  }
+
+  private static class TestFactory extends SamAuthenticatedUserRequestFactory {
```
pshapiro4broad: Subclassing is fine but I think using `spy()` to do this is a more idiomatic way to use the mocks library. For example: https://github.com/DataBiosphere/terra-data-catalog/blob/main/service/src/test/java/bio/terra/catalog/iam/SamServiceTest.java

pshapiro4broad (approved):
I have many comments but overall this seems fine. I've given a 👍  but please check the comments out before merging.

While most services should never need to check `enabled` themselves, I do worry that now SAM clients need to think about if they need to or not. But I see your point that every service will need to check at _some_ point if the user is authorized for an operation or not, and that will (typically) require talking to SAM.

dvoet:
> While most services should never need to check enabled themselves, I do worry that now SAM clients need to think about if they need to or not. But I see your point that every service will need to check at some point if the user is authorized for an operation or not, and that will (typically) require talking to SAM.

@pshapiro4broad I don't believe thinking is required. Almost all cases ask Sam if a user can do a thing or need to know who the user is in order to operate on some user-private resource. If your use cases are different then certainly follow the least error prone path.

src/main/java/bio/terra/common/iam/SamAuthenticatedUserRequestFactory.java
dvoet: I don't know why `SamUnauthorizedException` or all the other Sam wrapper exceptions exist.

src/main/java/bio/terra/common/iam/SamUserAuthenticatedRequest.java
dvoet: I was mainly trying to stick with the preexisting terminology. I agree the names are clunky.

src/main/java/bio/terra/common/iam/TokenAuthenticatedRequest.java
dvoet: I am just copying the style of the code in the area. I don't really know why builders are used in the first place here and even hand rolled. I was assuming stairway or something cares.

yonghaoy:
`It extracted the email address and user id from the access token which are not alway correct.`
[AuthenticatedUserRequest.java](https://github.com/DataBiosphere/terra-common-lib/pull/73/files#diff-be3cc64f504fc044a617539415a4313f7dde5f49383741fb5d56f8bcf51441bf) does not have this assumption. It just presents an authenticated user(e.g. already authenticated by proxy). 
[ProxiedAuthenticatedUserRequestFactory.java](https://github.com/DataBiosphere/terra-common-lib/pull/73/files#diff-3e9c8e640650e3b15b418b46ff11e054ebd97ee4f9bd8c21dd6a8d3effbb0e2d) has the assumption that it can extract email and user id from access token because it is built for the existing Terra's Apache proxy + SAM infra.  ProxiedAuthenticatedUserRequestFactory name is bit misleading and too general. I believe there can be Nginx/IstioJwtProxiedAuthenticaedUserFactory in the future if we change the way how Terra authenticated user then put user id/email in different headers. But factory pattern still works in this case.  

`Querying sam for every request is a waste of electricity and is bad for the environment as the majority of api calls don't need that detail.` Does AuthenticatedUserRequest make that query? I through that query is made by Apache proxy and the services behind proxy does not make the call. 

For the case that I want to use user email or id but I don't want to invoke SAM call, shall I still use `SamUserAuthenticatedRequest`?

dvoet:
> `It extracted the email address and user id from the access token which are not alway correct.`
> [AuthenticatedUserRequest.java](https://github.com/DataBiosphere/terra-common-lib/pull/73/files#diff-be3cc64f504fc044a617539415a4313f7dde5f49383741fb5d56f8bcf51441bf) does not have this assumption. It just presents an authenticated user(e.g. already authenticated by proxy).

The fact that `AuthenticatedUserRequest` contains user id and email as required fields is an assumption that they are always present. The way we use the existing factory is wrong but I don't want to blindly swap it out for a factory that calls sam when that information is not usually needed. Deprecating the old approach is a good forcing function.

> query is made by Apache proxy

This is part of ongoing work to [deprecate opendj](https://docs.google.com/document/d/1XTSne7Yd3hUsjK3yG3bNwaq9LzuDxxjMvmVEsAyYEP8/edit#heading=h.cdz6o4p358kc) and remove these calls from the proxy.

> For the case that I want to use user email or id but I don't want to invoke SAM call, shall I still use SamUserAuthenticatedRequest?

yes

yonghaoy:
> > `It extracted the email address and user id from the access token which are not alway correct.`
> > [AuthenticatedUserRequest.java](https://github.com/DataBiosphere/terra-common-lib/pull/73/files#diff-be3cc64f504fc044a617539415a4313f7dde5f49383741fb5d56f8bcf51441bf) does not have this assumption. It just presents an authenticated user(e.g. already authenticated by proxy).
> 
> The fact that `AuthenticatedUserRequest` contains user id and email as required fields is an assumption that they are always present. The way we use the existing factory is wrong but I don't want to blindly swap it out for a factory that calls sam when that information is not usually needed. Deprecating the old approach is a good forcing function.
> 
"a factory that calls sam when that information is not usually needed" what is the factory? I don't think ProxyAuthFactory makes SAM call. Is that a new factory we need to add after deprecating opendj?

> > query is made by Apache proxy
> 
> This is part of ongoing work to [deprecate opendj](https://docs.google.com/document/d/1XTSne7Yd3hUsjK3yG3bNwaq9LzuDxxjMvmVEsAyYEP8/edit#heading=h.cdz6o4p358kc) and remove these calls from the proxy.
I don't have enough context on this ongoing work(and I don't have access to that doc). After removing that doc, how would Terra force all services to verify user enabled status? Is the new principle: 1: If the service make SAM calls, then SAM will take care of this check 2: Otherwise, the service use SamUserAuthenticatedRequest to check user status themself? 

> 
> > For the case that I want to use user email or id but I don't want to invoke SAM call, shall I still use SamUserAuthenticatedRequest?
> 
> yes
In this case, I feel I'd like to use `ProxyAuthenticatedUserRequest`.

ddietterich:
> ... the majority of api calls don't need that detail ...
How does this relate to the plan to remove OpenDJ? If the proxy is not verifying that the incoming user is a registered Terra user, then I would think the majority of API calls will need to use the `SamUserAuthenticatedRequest` version. Do I understand that correctly?

dvoet:
@yonghaoy the `ProxyAuthenticatedUserRequest` may give wrong information because the email and id may be for a pet and not the user. That is why I deprecated it. Any code that relies on this method is open to bugs where pet service accounts cannot be used to make api calls.

@ddietterich 
>... the majority of api calls don't need that detail ...
How does this relate to the plan to remove OpenDJ? If the proxy is not verifying that the incoming user is a registered Terra user, then I would think the majority of API calls will need to use the SamUserAuthenticatedRequest version. Do I understand that correctly?

there are 2 main classes of requests: 1) performs some action and asks sam if the user is permitted to do that action or 2) performs some user-private action that does not require a sam check but needs to know who the user is. In the case of 1 the sam permission check will ensure the user is active. In the case of 2 the `SamUserAuthenticatedRequest` should be used.

This is related to the plan to remove OpenDJ by correcting existing patterns for getting a user's id and email and creating common code to ensure a user is enabled

src/main/java/bio/terra/common/iam/TokenAuthenticatedRequestFactory.java
```diff
@@ -0,0 +1,28 @@
...
+
+  public TokenAuthenticatedRequest from(HttpServletRequest servletRequest) {
+    TokenAuthenticatedRequest.Builder builder = TokenAuthenticatedRequest.builder();
+
+    String authHeader = servletRequest.getHeader(AUTHORIZATION);
```
rtitle: Technically we also get token expiration from the Apache proxy, but I don't know of any services which actually use that field -- so fine to leave it out I think.

rtitle (approved):
Looks good. I like how this code forces you to do the right thing -- e.g. if you really need the user email then you need to choose to call Sam; otherwise you just get the token.

src/main/java/bio/terra/common/iam/TokenAuthenticatedRequestFactory.java
dvoet: I don't think expiration is ever used. The preexisting code did not get it either.

ddietterich: Is there a way to get the expiration after the fact if we need it?

src/main/java/bio/terra/common/iam/SamAuthenticatedUserRequestFactory.java
ddietterich: IIRC the Sam exceptions don't match the TCL base exception, so they are wrapped for consistency upstream.

src/main/java/bio/terra/common/iam/SamAuthenticatedUserRequestFactory.java
ddietterich: Yes, the object handling here is odd. What do you expect as the lifespan of that client? I think the caller would need to create one for each call to this, yes? It couldn't be a shared object because concurrent requests would mess each other up.

If you don't have a way to make your own client object, then please document the expectations of that input.

src/main/java/bio/terra/common/iam/SamUserAuthenticatedRequest.java
ddietterich: I thought `request` was there because the information for creating the object came from the HTTP request object. I would not object to different names, but the ones you have are fine with me.

It does look to me like you have mismatched names between the factory and the object.

ddietterich (changes requested):
This looks good. Changes I would like to see:
1. Change or documentation around the use of SamApiClient
2. Fixing naming: either SamAuthenticatedUserRequest or SamUserAuthenticatedRequest, but not both. Or as mentioned in the comments, something else entirely.

src/main/java/bio/terra/common/iam/SamAuthenticatedUserRequestFactory.java
dvoet: I think I will just pass in the path

src/main/java/bio/terra/common/iam/TokenAuthenticatedRequestFactory.java
dvoet: it is in the header if we need it. In all the years of FireCloud's and Terra's existence it has been there and we have never used it.

ddietterich: SGTM

dvoet:
@pshapiro4broad @rtitle @ddietterich I have substantially refactored, PTAL. 
- Renamed everything
- Removed builders
- Addressed most feedback

ddietterich (approved):
LGTM

src/test/java/bio/terra/common/iam/SamUserFactoryTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+  private static final SamUser SAM_USER =
+      new SamUser("test@example.com", "Subject", new BearerToken("0123.456-789AbCd"));
+  private static final String SAM_BASE_PATH = "not_real";
+
+  @Test
+  public void enabledUser() throws ApiException {
+    SamUserFactory factory = spy(new SamUserFactory(new BearerTokenFactory()));
+    UsersApi usersApi = mock(UsersApi.class);
+    when(factory.createUsersApi(SAM_USER.getBearerToken(), SAM_BASE_PATH)).thenReturn(usersApi);
```
pshapiro4broad, line 25: The main downside to `spy()` is that mocking a spied object method is different than mocking a mocked object method. Writing the mock this way will result in a call to to the real object's `factory.createUsersApi()` method. Instead you need to mock using this form:
```
    doReturn(usersApi).when(factory).createUsersApi(SAM_USER.getBearerToken(), SAM_BASE_PATH);
```

PR #74, opened 2022-06-23
[PF-548] Add okhttp support and move to b3 format trace headers

zloery:
This has a couple changes to support better (and cross-service!) tracing in GCP:

- Move the common-lib tracing code to read B3 headers from incoming requests instead of the default traceContext format. Sam only supports B3 headers, so we should prefer using that everywhere instead of mixing formats across services.
- Fix the stairway tracing hook to properly track span parents
- Add an interceptor for OkHttp-based client libraries. TCL users can add this interceptor to an okHttp client library (like the Sam Java client) to propagate traces across services.

### DIFF
--- a/src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
+++ b/src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
@@ -10,11 +10,18 @@ import okhttp3.Interceptor;
 import okhttp3.Request;
 import okhttp3.Response;
 
+/**
+ * An interceptor to add tracing headers to outgoing requests.
+ *
+ * OkHttp interceptors are called before and after network requests, and can modify the request and
+ * response objects. In this case, we're adding headers for the current traceId and span, which the
+ * remote service can read to link traces across services.
+ */
 public class OkHttpClientTracingInterceptor implements Interceptor {
-  private Tracer tracer;
+  private final Tracer tracer;
   // This looks a little odd, but OkHttp requests are immutable so the carrier type must be a
   // request builder, not a Request.
-  private HttpClientHandler<Request, Response, Request.Builder> handler;
+  private final HttpClientHandler<Request, Response, Request.Builder> handler;
   private static final Setter<Request.Builder> SETTER =
       new Setter<Request.Builder>() {
         @Override
--- a/src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
+++ b/src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
@@ -13,9 +13,9 @@ import okhttp3.Response;
 /**
  * An interceptor to add tracing headers to outgoing requests.
  *
- * OkHttp interceptors are called before and after network requests, and can modify the request and
- * response objects. In this case, we're adding headers for the current traceId and span, which the
- * remote service can read to link traces across services.
+ * <p>OkHttp interceptors are called before and after network requests, and can modify the request
+ * and response objects. In this case, we're adding headers for the current traceId and span, which
+ * the remote service can read to link traces across services.
  */
 public class OkHttpClientTracingInterceptor implements Interceptor {
   private final Tracer tracer;
--- a/src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
+++ b/src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
@@ -1,0 +1,56 @@
+package bio.terra.common.tracing;
+
+import io.opencensus.contrib.http.HttpClientHandler;
+import io.opencensus.contrib.http.HttpRequestContext;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import io.opencensus.trace.propagation.TextFormat.Setter;
+import java.io.IOException;
+import okhttp3.Interceptor;
+import okhttp3.Request;
+import okhttp3.Response;
+
+public class OkHttpClientTracingInterceptor implements Interceptor {
+  private Tracer tracer;
+  // This looks a little odd, but OkHttp requests are immutable so the carrier type must be a
+  // request builder, not a Request.
+  private HttpClientHandler<Request, Response, Request.Builder> handler;
+  private static final Setter<Request.Builder> SETTER =
+      new Setter<Request.Builder>() {
+        @Override
+        public void put(Request.Builder carrier, String key, String value) {
+          carrier.header(key, value);
+        }
+      };
+
+  public OkHttpClientTracingInterceptor(Tracer tracer) {
+    this.tracer = tracer;
+    this.handler =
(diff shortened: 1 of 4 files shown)


jaycarlton:
At risk of sounding naive, did you estimate the cost of making Sam more flexible in its header handling?

src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
```diff
@@ -0,0 +1,56 @@
...
+import io.opencensus.contrib.http.HttpRequestContext;
+import io.opencensus.trace.Tracer;
+import io.opencensus.trace.Tracing;
+import io.opencensus.trace.propagation.TextFormat.Setter;
+import java.io.IOException;
+import okhttp3.Interceptor;
+import okhttp3.Request;
+import okhttp3.Response;
+
+public class OkHttpClientTracingInterceptor implements Interceptor {
```
jaycarlton, line 20: Put a little Javadoc here about what this is for.

src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
```diff
@@ -0,0 +1,56 @@
...
+  private Tracer tracer;
+  // This looks a little odd, but OkHttp requests are immutable so the carrier type must be a
+  // request builder, not a Request.
+  private HttpClientHandler<Request, Response, Request.Builder> handler;
+  private static final Setter<Request.Builder> SETTER =
+      new Setter<Request.Builder>() {
+        @Override
```
jaycarlton, line 27: can this be a lambda? I.e. is it a functional interface? Maybe not if there's not actually an interface. 🤔

src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
```diff
@@ -0,0 +1,56 @@
...
+    HttpRequestContext requestContext =
+        handler.handleStart(tracer.getCurrentSpan(), newRequestBuilder, originalRequest);
+    Request newRequest = newRequestBuilder.build();
+    Response response = null;
+    try {
+      response = chain.proceed(newRequest);
+    } catch (IOException e) {
+      // Error during call
+      handler.handleEnd(requestContext, newRequest, null, e);
+      throw e;
```
jaycarlton, line 57: Where will this exception wind up? Should you add a message?

src/main/java/bio/terra/common/tracing/OkHttpTracingExtractor.java
```diff
@@ -0,0 +1,53 @@
...
+import okhttp3.Response;
+
+public class OkHttpTracingExtractor extends HttpExtractor<Request, Response> {
+
+  @Nullable
+  @Override
+  public String getRoute(Request request) {
+    // OpenCensus spec wants this to be something like the literal string
+    // "/api/workspace/{workspaceId}", whereas path would be "/api/workspace/12345".
+    // We don't have the route available and this is a best-effort method, so use path instead.
```
jaycarlton: Does this mean we'll get many more span categories than we would otherwise? How hard is it to aggregate across these?

What happens if you return `null` or an empty string for the route?

src/main/java/bio/terra/common/tracing/OkHttpTracingExtractor.java
```diff
@@ -0,0 +1,53 @@
...
+
+  @Nullable
+  @Override
+  public String getUserAgent(Request request) {
+    return request.header("user-agent");
+  }
+
+  @Override
+  public int getStatusCode(@Nullable Response response) {
+    return response != null ? response.code() : 0;
```
jaycarlton, line 59: Is this supposed to be an HTTP response code? Maybe returning -1 would be less ambiguous, though I guess 0 isn't a valid response code either.

jaycarlton (approved):

zloery:
> At risk of sounding naive, did you estimate the cost of making Sam more flexible in its header handling?

Fair question, I only took a quick look into it before starting on this. I think it's "moderate but non-trivial work":
- The scala wrapper for OpenCensus that Sam is using (https://github.com/census-ecosystem/opencensus-scala) only supports B3 format. I don't think the traceContext format is particularly complicated, but you'd have to add it to that library, and the owners put it in archive mode 2 years ago so they wouldn't be adding it themselves.
- I don't know if there are other scala wrappers for OpenCensus, but swapping them in Sam would be a moderate lift as well. Given that we're planning to migrate to OpenTelemetry (soon™️ ), it's probably not worth swapping out the OpenCensus implementation.

jaycarlton:
> Given that we're planning to migrate to OpenTelemetry (soon™️ ), it's probably not worth swapping out the OpenCensus implementation.

I agree. Thanks.

src/main/java/bio/terra/common/tracing/OkHttpTracingExtractor.java
zloery: This is normally an HTTP status code, though the doc on the method I'm implementing explicitly requires "If the response is null, this method should return 0". Added that as a comment as well since it's a little strange

src/main/java/bio/terra/common/tracing/OkHttpTracingExtractor.java
zloery: These methods just populate the fields of a span which will later show up in GCP, e.g. this image. All fields are nullable here, so if you return `null` it just won't populate that field of the span. IMO either is fine, thinking about it more it's probably better to leave this blank than put a slightly-wrong value.

src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
zloery: Interceptors wrap the actual HTTP request, so throwing here means the exception goes up to the previous interceptor in the chain (and eventually to the client library that made the actual call). 
Since this interceptor is only handling tracing and this exception is coming from the call, I think it's better to pass the exception without modifying it here.

src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
zloery: I don't think so because `Setter` is an abstract class and not an interface, but I'm far from sure on that.

src/main/java/bio/terra/common/tracing/OkHttpTracingExtractor.java
jaycarlton: I'd probably invert the ternary so that it's one less backflip for the brain.

src/main/java/bio/terra/common/tracing/OkHttpClientTracingInterceptor.java
jaycarlton: IJ will highlight the section if it's lambda-able.

ddietterich (approved):
LGTM

PR #78, opened 2022-08-08
Upgrade liquibase (WOR-441).

cahrens:
This is related to https://github.com/DataBiosphere/terra-workspace-manager/pull/753. I naively hand-edited the integration lockfile in that PR, and now that I'm actually running `/gradlew :integration:dependencies --write-locks`, I see that `terra-common-lib` is causing the older liquibase version to be used in the integration directory (only).


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.0-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.1-bin.zip
```
cahrens, line 3: Without this, I get an error when running gradle due to incompatible JDK/Gradle versions. I'm upgrading to the same version that is used in WSM.

cahrens:
I was able to test this by publishing to my local maven repository, then in my WSM config I set it to use the local maven repository. I then ran `./gradlew :integration:dependencies --write-locks` and verified that the postres and liquibase versions remained the upgraded versions (though plenty of other things changed, so it was definitely picking up the modified `terra-common-lib` library.

blakery (approved):

marctalbott (approved):

PR #80, opened 2022-08-09
Update spring dependencies (WOR-433).

cahrens:


cahrens:
Punting on this, per comment in WSM PR #762.

PR #84, opened 2022-09-09
[PF-1959] Add DataSourceManager; refactor DataSourceInitializer

ddietterich:
`DataSourceInitializer` provides a static method that creates a connection pooled data source with settings taken from the database settings. It has a very specific problem: if you are running JUnit tests and you use the `@DirtiesContext` annotation, then Spring will destroy the associated application context. In the process, it abandons the connection pools without closing the connections. In the WSM case that turned up this problem, we ran out of database connections.

The new `DataSourceManager` is a `@Component` that does the same thing as `DataSourceInitializer` (I set it up to share the pool construction code) but the Manager remembers the pools that it has created. It supplies an `@PreDestroy` method that closes the pools, and thus the connections, when the application context is destroyed.

This has no effect in production; when the container exits, the whole JVM goes away and all of the database connections with it.

@nmalfroy I thought you might be particularly interested for TDR testing

### DIFF
--- a/src/main/java/bio/terra/common/db/DataSourceManager.java
+++ b/src/main/java/bio/terra/common/db/DataSourceManager.java
@@ -1,51 +1,44 @@
 package bio.terra.common.db;
 
-import org.apache.commons.dbcp2.ConnectionFactory;
-import org.apache.commons.dbcp2.DriverManagerConnectionFactory;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.PreDestroy;
+import javax.sql.DataSource;
 import org.apache.commons.dbcp2.PoolableConnection;
-import org.apache.commons.dbcp2.PoolableConnectionFactory;
 import org.apache.commons.dbcp2.PoolingDataSource;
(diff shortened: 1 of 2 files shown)


nmalfroy:
Interesting, thanks for pinging me!  I'll try to take a look today or Monday

mikenev (approved):
Interesting issue.

src/main/java/bio/terra/common/db/DataSourceManager.java
```diff
@@ -0,0 +1,44 @@
...
+  private List<ObjectPool<PoolableConnection>> connectionPools = new ArrayList<>();
+
+  @PreDestroy
+  public void destroy() {
+    logger.info("Closing all connection pools");
+    for (var pool : connectionPools) {
+      logger.info("Pool active {} idle {}", pool.getNumActive(), pool.getNumIdle());
+      pool.close();
+    }
+    connectionPools = new ArrayList<>();
```
nmalfroy, line 35: Is there any value in logging how many pools will be deleted?

nmalfroy (approved):
looks good!

PR #86, opened 2022-09-19
[PF-2016] Stop logging auth headers

zloery:
While debugging a different issue I noticed we were including access tokens in logs, which we shouldn't do.


yuhuyoyo (approved):

ddietterich (approved):
LGTM - I'll trust that those are the right strings to look for!

PR #92, opened 2022-11-17
[DC-420] Add gatewaytimeoutexception

s-rubenstein:


pshapiro4broad (approved):
The change looks fine, but I wonder if 504 is the correct code to use in a service that's not a proxy or gateway.

From reading https://www.rfc-editor.org/rfc/rfc7231#section-6.6.5 it sounds like 504 is intended to be used for a service that's acting as a gateway or proxy. So it wouldn't be used in the case where one service calls another as part of its operation.

s-rubenstein:
@pshapiro4broad the other code I'd been thinking about was https://www.rfc-editor.org/rfc/rfc7231#section-6.5.7. Do you think that would more accurately capture the issue?

pshapiro4broad:
> @pshapiro4broad the other code I'd been thinking about was https://www.rfc-editor.org/rfc/rfc7231#section-6.5.7. Do you think that would more accurately capture the issue?

That sounds like the timeout occurred while the server was waiting for the request. Since 4xx are client errors and 5xx are server errors, a 5xx error makes more sense to me. I think 500 may be the best error. The server encountered an unexpected error while performing an operation that was otherwise correct.

PR #95, opened 2023-01-19
[PF-2421] Update dependency / library versions

dexamundsen:
- Upgrade dependency versions
- Clean up / standardize build files to keep them consistent
- required in https://github.com/DataBiosphere/terra-workspace-manager/pull/1012


dexamundsen:
will be handled downstream after private fork

PR #104, opened 2023-04-12
Add bump-skip check

dexamundsen:
Current - bumper bumps the version on github, this is not identified as a bump, and hence a second artifact is created and uploaded. Since the second artifact uses the same semver version, this does not continue endlessly. (bumper sees current version same as new version and hence does not commit another change to github, thus breaking the cycle)

Fix - add bum-skip check

Also adding fix from - https://github.com/DataBiosphere/terra-cloud-resource-lib/pull/163

image


.github/actions/setup-gradlew /action.yml
```diff
@@ -0,0 +1,28 @@
+# Copied from https://github.com/DataBiosphere/terra-aws-resource-discovery/tree/main
+# TODO(TERRA-418) Move this to https://github.com/DataBiosphere/github-actions
+# This action must be done after the checkout action
+name: 'setup-gradlew'
```
zloery: nit: this directory name has a space after `setup-gradlew` that it probably shouldn't

.github/workflows/develop-push.yml
```diff
@@ -19,54 +19,54 @@ jobs:
...
+
+    - name: Build, assemble, and test
+      id: build-assemble-test
+      run: ./gradlew build
+    - name: Upload Test Reports
+      if: always()
+      uses: actions/upload-artifact@v3
+      with:
+        name: Test Reports
+        path: build/reports
```
zloery, line 53: we can also skip building and testing on version bumps, those will be covered by the main PR

zloery (approved):

.github/actions/setup-gradlew /action.yml
dexamundsen: Thanks, done

.github/workflows/develop-push.yml
dexamundsen: I am inclined to keep it 
1) Unlikely for TCL, but if we have second PR get merged w/o any conflicts and w/o a rebase - it is possible that the second PR broke functionality. Keeping this test will help us identify the breaking-change, and prevent a broken artifact from being published
2) Libraries dont have nightly tests, so these may as well function as post-merge sanity tests

yuhuyoyo (approved):

PR #105, opened 2023-04-17
PF-2692: add a flagsmith service

yuhuyoyo:
A small wrapper around the java library


src/test/resources/application-flagsmith-test.properties
```diff
@@ -0,0 +1,3 @@
+terra.common.flagsmith.enabled=true
+terra.common.flagsmith.server-side-api-key=ser.Vn8MpGpej26aShW7CXsEaR
+terra.common.flagsmith.api-url=https://terra-flagsmith.api.verily.com/api/v1/
```
zloery: Are these ok to check into terra-common-lib? And is it ok for this upstream repo to call the internal-env instance?

yuhuyoyo: i created a test env but now I moved it to use the public one.

zloery (approved):

src/main/java/bio/terra/common/flagsmith/FlagsmithService.java
```diff
@@ -0,0 +1,50 @@
...
+public class FlagsmithService {
+  private static final Logger LOGGER = LoggerFactory.getLogger(FlagsmithService.class);
+
+  @Autowired private final FlagsmithProperties flagsmithProperties;
+
+  FlagsmithService(FlagsmithProperties flagsmithProperties) {
+    this.flagsmithProperties = flagsmithProperties;
+  }
+
+  public boolean isFeatureEnabled(String feature, Optional<Boolean> defaultValue) {
```
dexamundsen: Minor: consider this signature

public Optional<Boolean> isFeatureEnabled(String feature) {
  // if flagsmith is not enabled, optional.emtpy()
}

reason : this library does not need to know the default value. Default behavior is determined by the client. 

Alternate - throw an exception if flagsmith is not set up.

dexamundsen (approved):

src/main/java/bio/terra/common/flagsmith/FlagsmithService.java
yuhuyoyo: good idea.

PR #106, opened 2023-04-18
[PF-2697] Use new version of Stairway

ddietterich:
I also fixed the database initialization script to grant permissions to the tclstairwayuser. Otherwise, the liquibase steps do not have permission to run.

### DIFF
--- a/local-dev/local-postgres-init.sql
+++ b/local-dev/local-postgres-init.sql
@@ -2,3 +2,6 @@ CREATE DATABASE testdb;
 CREATE ROLE dbuser WITH LOGIN ENCRYPTED PASSWORD 'dbpwd';
 CREATE DATABASE tclstairway;
 CREATE ROLE tclstairwayuser WITH LOGIN ENCRYPTED PASSWORD 'tclstairwaypwd';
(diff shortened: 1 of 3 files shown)


local-dev/local-postgres-init.sql
```diff
@@ -2,3 +2,6 @@ CREATE DATABASE testdb;
 CREATE ROLE dbuser WITH LOGIN ENCRYPTED PASSWORD 'dbpwd';
 CREATE DATABASE tclstairway;
 CREATE ROLE tclstairwayuser WITH LOGIN ENCRYPTED PASSWORD 'tclstairwaypwd';
+GRANT ALL ON DATABASE tclstairway TO tclstairwayuser WITH GRANT OPTION;
```
yuhuyoyo, line 5: what is this for?

yuhuyoyo (approved):

dexamundsen (approved):

local-dev/local-postgres-init.sql
ddietterich: I used the script to create a new database. The tests failed because the tclstairwayuser was not given access to the database or to the public schema.

PR #108, opened 2023-05-10
PF-2751: add retry and add more cache config

yuhuyoyo:
retry when failed to get the list of flags in a given environment for up to 1 min.

### DIFF
--- a/src/main/java/bio/terra/common/flagsmith/FlagsmithService.java
+++ b/src/main/java/bio/terra/common/flagsmith/FlagsmithService.java
@@ -4,7 +4,11 @@ import com.flagsmith.FlagsmithClient;
 import com.flagsmith.config.FlagsmithCacheConfig;
 import com.flagsmith.exceptions.FlagsmithClientError;
 import com.flagsmith.models.Flags;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/common/flagsmith/FlagsmithService.java
```diff
@@ -32,25 +38,62 @@ public Optional<Boolean> isFeatureEnabled(String feature) {
...
-      LOGGER.warn("Failed to get the feature state of {}", feature, e);
-      return Optional.empty();
+      LOGGER.warn("Feature {} not found in {}", feature, flagsmithProperties.getApiUrl(), e);
+    } catch (Exception e) {
+      LOGGER.warn("Failed to fetch feature {}", feature, e);
+    }
+    return Optional.empty();
+  }
+
+  private static <T> T getWithRetryOnException(SupplierWithException<T> supplier) throws Exception {
```
ddietterich, line 64: I should move the RetryUtils into common lib... oh well.

ddietterich (approved):

src/main/java/bio/terra/common/flagsmith/FlagsmithService.java
yuhuyoyo: yeah that would have been really nice :)

jmczerk (approved):

PR #110, opened 2023-05-24
PF-2809: Create custom metrics for stairway flight to record latency and flight errors

yuhuyoyo:


ddietterich (approved):
Cool! Seems like a reasonable start for flight metrics. I could see splitting out the FATAL from the ERROR counts, since FATAL are dismal failures.

yuhuyoyo:
> Cool! Seems like a reasonable start for flight metrics. I could see splitting out the FATAL from the ERROR counts, since FATAL are dismal failures.

you can group them by status. so you can see the distribution of error vs fatal in that metrics.

yonghaoy (approved):
Change LG. My only concern is the name of the Hook. With metrics support, it is nolonger 
`tracingHook` anymore as it does something extra and becomes a general `monitoring hook`.
Not a big concern, feel free to file a follow up ticket if the feedback makes sense..

yonghaoy (approved):
Change LG. My only concern is the name of the Hook. With metrics support, it is nolonger 
`tracingHook` anymore as it does something extra and becomes a general `monitoring hook`.
Not a big concern, feel free to file a follow up ticket if the feedback makes sense..

yonghaoy (approved):
And a nice & more useful improvement could be Flight step metrics. Knowing how long each steps takes and state would be useful.

yuhuyoyo:
> And a nice & more useful improvement could be Flight step metrics. Knowing how long each steps takes and state would be useful.

agree. but i will do it as incremental update.

PR #111, opened 2023-05-26
PF-2809: add step latency metrics

yuhuyoyo:
I was going to let the flight latency metrics soaks for a bit but i realize I need to rename the path of that so here's the step latency metrics as well.


yuhuyoyo:
i have confirmed that the custom metrics works and is showing up in devel env.

ddietterich (approved):

yonghaoy (approved):
LG. One thing to consider is the latency might be vary depends on the status. flight/step status might be a good dimension?

PR #112, opened 2023-05-26
PF-2813: Add a method to retrieve flagsmith feature value

yuhuyoyo:


src/main/java/bio/terra/common/flagsmith/FlagsmithService.java
```diff
@@ -61,6 +54,47 @@ public Optional<Boolean> isFeatureEnabled(String feature) {
...
+    try {
+      Flags flags = getWithRetryOnException(flagsmith::getEnvironmentFlags);
+      Object value = flags.getFeatureValue(feature);
+      if (value == null) {
+        return Optional.empty();
+      }
+      return Optional.of(objectMapper.readValue(value.toString(), clazz));
+    } catch (FlagsmithClientError e) {
+      LOGGER.warn("Feature {} not found in {}", feature, flagsmithProperties.getApiUrl(), e);
+    } catch (Exception e) {
```
ddietterich, line 83: If `getWithRetryOnException` throws `InterruptedException` this will swallow that error.
I think it should be allowed to propagate.

ddietterich (commented):
Interrupt handling looks sketchy.

ddietterich (approved):

src/main/java/bio/terra/common/flagsmith/FlagsmithService.java
```diff
@@ -57,10 +50,53 @@ public Optional<Boolean> isFeatureEnabled(String feature) {
...
     return Optional.empty();
   }
 
+  /**
+   * Get feature's value formatted as JSON.
+   *
+   * <p>If Flagsmith is unavailable, feature does not exist or the feature value does not exist,
+   * return {@code Optional.empty()}.
+   */
+  public <T> Optional<T> getFeatureValueJson(String feature, Class<T> clazz) throws Exception {
```
dexamundsen: nit: should this just be named getFeatureValue? since the type is controlled by clazz

src/main/java/bio/terra/common/flagsmith/FlagsmithService.java
```diff
@@ -57,10 +50,53 @@ public Optional<Boolean> isFeatureEnabled(String feature) {
       LOGGER.warn("Feature {} not found in {}", feature, flagsmithProperties.getApiUrl(), e);
     } catch (Exception e) {
       LOGGER.warn("Failed to fetch feature {}", feature, e);
+      throw e;
```
dexamundsen: minor: may be throw a specific type of exception to help ops?

dexamundsen (approved):
minor comment, not blockers

src/main/java/bio/terra/common/flagsmith/FlagsmithService.java
yuhuyoyo: objectMapper.readValue will throw if it's not a json value. the flagsmith value doesn't required to be json.

PR #113, opened 2023-06-09
Update io.kubernetes for jose4j vulnerability (WOR-1040).

cahrens:
We are seeing in Rawls that `jose4j` 0.9.2 is being brought in through `io.kubernetes` 17.0 via `terra-common-lib` 0.91.SNAPSHOT. 

image

I see that @zloery pinned jose4j in https://github.com/DataBiosphere/terra-common-lib/pull/109, but did not upgrade `io.kubernetes`. When I upgrade `io.kubernetes`, it updates `jose4j` and we can remove the pinning.


build.gradle
```diff
@@ -120,7 +120,6 @@ dependencies {
     constraints {
         implementation('org.scala-lang:scala-library:2.13.10')
         implementation('org.json:json:20230227')
-        implementation('org.bitbucket.b_c:jose4j:0.9.3')
```
cahrens, line 123: Can remove this and it remains [at 0.9.3](https://github.com/DataBiosphere/terra-common-lib/blob/c1724acb41ca2dbe3f9e9e96157b3eaf58350a1f/gradle.lockfile#L157) after executing --write-locks.

cahrens:
Ugh, but I see the comment from @zloery : "This is brought in via io.kubernetes:client-java, but newer versions of that library have incompatible dependencies with Spring until we move to Spring Boot 3". So we can't upgrade `io-kubernetes` to version 18? :(

zloery:
> Ugh, but I see the comment from @zloery : "This is brought in via io.kubernetes:client-java, but newer versions of that library have incompatible dependencies with Spring until we move to Spring Boot 3". So we can't upgrade io-kubernetes to version 18? :(

Yeah, that was the issue I hit doing this upgrade before - I think the conflicting libraries were `org.yaml:snakeyaml` and `org.slf4j:slf4j-api`. Spring isn't happy with the new major versions that kubernetes client-java pulls in, but it might be possible to pin those libraries to the current versions if `io-kubernetes` 18 is still compatible with them. I prefer avoiding strict version constraints where possible so I didn't experiment with that much, but that might be worth trying if you can't get Rawls to use the right version otherwise.

This is also probably a good nudge on upgrading our services to spring boot 3, I'll check in on the current state of that

PR #117, opened 2023-08-31
[BENCH-1050] reduce cache expiry time in flagsmith

dexamundsen:
reduce cache expiry time in flagsmithservice from 12 hrs to 10 min
This will be used to in CLI as well 
Update javadoc comments


src/main/java/bio/terra/common/flagsmith/FlagsmithClient.java
```diff
@@ -0,0 +1,101 @@
...
+import com.flagsmith.models.Flags;
+import java.util.Map;
+import java.util.Optional;
+import javax.annotation.Nullable;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+
+public class FlagsmithClient {
+  private static final Logger LOGGER = getLogger(FlagsmithClient.class);
+  private final com.flagsmith.FlagsmithClient flagsmith;
```
rogerwangcs: why not just import `FlagsmithClient`?

src/main/java/bio/terra/common/flagsmith/FlagsmithClient.java
```diff
@@ -0,0 +1,101 @@
...
+
+  private Flags getFlags(String userEmail) throws FlagsmithClientError {
+    if (userEmail != null) {
+      flagsmith.getIdentityFlags(userEmail, Map.of("email_address", userEmail));
+    }
+    return flagsmith.getEnvironmentFlags();
+  }
+
+  // list of features
+  public enum Features {
```
rogerwangcs: It seems pretty painful that we need to maintain the set of features in TCL. Are there any drawbacks to letting consumers manage the subset of flags they need and pass in string?

src/main/java/bio/terra/common/flagsmith/FlagsmithClient.java
dexamundsen: this library wraps the identity traits we want to query against, example - email_address. 
i would anticipate we use more traits in the future. 

adding this wrapper will reduce the maintenance overhead (avoids repeating code in multiple repos)

rogerwangcs: Ah to clarify, I meant why not just import `FlagsmithClient` above and have `private final FlagsmithClient flagsmith;` here

src/main/java/bio/terra/common/flagsmith/FlagsmithClient.java
dexamundsen: +1 to this. 
Agree that it is an overhead. I debated it myself and here is the reason i stuck to it - 

- consolidating the list of features here: we dont need to hard code it in multiple repos. lettng consumers pass it as a param: hard coded values to be added in multiple repos, and is the value is incorrect a default 'false' is returned
- ideally this is done only once per major project/feature: for example , add it only once for AWS. hence the overhead is minimal
- With monorepo above overhead will further be reduced 

Not strongly opinionated on using enum though

rogerwangcs (approved):

src/main/java/bio/terra/common/flagsmith/FlagsmithClient.java
dexamundsen: Ah. gotcha. readability. just so we dont confuse it to a reference to the client class we are creating

src/main/java/bio/terra/common/flagsmith/FlagsmithClient.java
```diff
@@ -0,0 +1,101 @@
+package bio.terra.common.flagsmith;
+
```
yuhuyoyo: i don't understand why you are adding this. there is a flagsmithService that's doing the exact same thing already.

yuhuyoyo:
if you really want to, you can import TCL to CLI instead.

src/main/java/bio/terra/common/flagsmith/FlagsmithClient.java
dexamundsen: This is the bare minimum needed for the client, the service adds caching and many other features 

We don't want caching, hence this file

yuhuyoyo: i'm blocking the change because I think we need general consensus on caching. everywhere else is doing caching including UI.

dexamundsen:
> if you really want to, you can import TCL to CLI instead. 

Cannot use caching in cli. That would be a security concern. This is exactly like SAM calls in WSM - we make a call to SAM every time we need to check permission. We do not cache any SAM access or permission information.

src/main/java/bio/terra/common/flagsmith/FlagsmithClient.java
yuhuyoyo: if you want, you can make the caching time configurable. but there's no need to *duplicate* code

dexamundsen: I will set up a meeting, this PR does not change existing behavior in WSM/UI/Other and blocking a change is not needed. again we should not cache access information. 
I will set up a meeting if needed to resolve this. let me know.

src/main/java/bio/terra/common/flagsmith/FlagsmithClient.java
dexamundsen: removed since the service code is changed to be used in CLI.

src/main/java/bio/terra/common/flagsmith/FlagsmithClient.java
dexamundsen: Thanks for meeting, As discussed - reduced the cache expiry. Same client API key available in CLI can be used with service as well. hence redid the PR

mikenev (approved):

yuhuyoyo (approved):

PR #118, opened 2023-10-24
check if flag is enabled prior to checking flag value.

yuhuyoyo:


dexamundsen (approved):

src/main/java/bio/terra/common/flagsmith/FlagsmithService.java
```diff
@@ -93,6 +93,10 @@ public <T> Optional<T> getFeatureValueJson(
       LOGGER.info("Flagsmith is not enabled, use default value");
       return Optional.empty();
     }
+    if (!isFeatureEnabled(feature, userEmail).orElse(false)) {
```
zloery: This is making a second call to flagsmith that doesn't seem necessary. Can you make this check below on `flags` instead?

zloery (approved):

PR #119, opened 2023-10-27
WOR-1325 upgrade to spring 3, jakarta and open telemetry

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-1325

see https://github.com/DataBiosphere/terra-java-project-template/pull/21 for associated changes to the template project


settings.gradle
```diff
@@ -1,4 +1,4 @@
 rootProject.name = 'terra-common-lib'
 
 gradle.ext.projectGroup = 'bio.terra'
-gradle.ext.tclVersion = '0.0.95-SNAPSHOT'
+gradle.ext.tclVersion = '0.1.95-SNAPSHOT'
```
cahrens: Shouldn't this be 0.1.0?

src/main/java/bio/terra/common/iam/SamUserFactory.java
```diff
@@ -28,16 +29,16 @@
...
+  final OkHttpClient httpClient;
 
-  @Autowired
-  public SamUserFactory(BearerTokenFactory bearerTokenFactory) {
+  @Autowired(required = false)
+  public SamUserFactory(
+      BearerTokenFactory bearerTokenFactory, Optional<OpenTelemetry> openTelemetry) {
     this.bearerTokenFactory = bearerTokenFactory;
+    var apiClientBuilder = new ApiClient().getHttpClient().newBuilder();
+    openTelemetry.ifPresent(
```
cahrens: When do we expect that openTelemtry won't be present?

src/test/java/bio/terra/common/tracing/TracingTestApplication.java
```diff
@@ -12,7 +12,7 @@
     },
     // Scan the common tracing package to pick up tracing auto configuration. Scan the logging
```
cahrens, line 13: Should update the comment.

src/test/resources/application-tracing-test.properties
```diff
@@ -1,4 +1 @@
-terra.common.tracing.samplingRate: 1.0
-# Empty list of urlPatterns to allow all URLs to be traced.
-terra.common.tracing.urlPatterns:
```
cahrens, line 3: Since urlPatterns is no longer a setting, does this mean all URLs are traced?

cahrens:
Thank you so much for taking this upgrade on, @dvoet!!

I skimmed through the code and left a few questions.

Can you also provide a description of what testing you have done? For instance, do the unit tests provide enough coverage that you are fairly confident everything is working? Did you test this new version with any of our existing services?

src/test/resources/application-tracing-test.properties
dvoet: all urls are traced

src/main/java/bio/terra/common/iam/SamUserFactory.java
dvoet: this is a library, I was trying not to force anything

dvoet:
> Can you also provide a description of what testing you have done? For instance, do the unit tests provide enough coverage that you are fairly confident everything is working? Did you test this new version with any of our existing services?

https://github.com/DataBiosphere/terra-java-project-template/pull/21

src/main/java/bio/terra/common/tracing/JerseyTracingFilter.java
```diff
@@ -1,51 +1,55 @@
...
   @Override
   public void filter(ClientRequestContext requestContext) throws IOException {
-    opencensusRequestContext =
-        handler.handleStart(tracer.getCurrentSpan(), requestContext, requestContext);
+    this.requestSpan =
+        tracer
+            .spanBuilder(
+                "Sent.%s %s"
+                    .formatted(requestContext.getMethod(), requestContext.getUri().toString()))
+            .setSpanKind(SpanKind.SERVER)
```
zloery: If this class is a filter for Jersey-based client libraries, I think this should be SpanKind.CLIENT

src/main/java/bio/terra/common/tracing/TracingConfig.java
```diff
@@ -1,151 +1,36 @@
...
-    // format.
-    servletContext.setAttribute(
-        OcHttpServletFilter.OC_TRACE_PROPAGATOR, Tracing.getPropagationComponent().getB3Format());
-    FilterRegistrationBean<OcHttpServletFilter> registration =
-        new FilterRegistrationBean<>(new OcHttpServletFilter());
-    registration.setUrlPatterns(tracingProperties.getUrlPatterns());
-    return registration;
+  @Primary
+  @ConditionalOnProperty(
+      name = "terra.common.tracing.stackdriverExportEnabled",
```
zloery, line 20: `stackdriverExportEnabled` was deleted in TracingProperties, why include this here?

src/main/java/bio/terra/common/tracing/TracingProperties.java
```diff
@@ -1,50 +0,0 @@
...
-import org.springframework.boot.context.properties.ConfigurationProperties;
-
-/**
- * Defines configuration properties for the tracing package; see {@link TracingConfig} for their
- * use.
- */
-@ConfigurationProperties(prefix = "terra.common.tracing")
-public class TracingProperties {
-  /** Rate of trace sampling, 0.0 - 1.0 */
-  private double samplingRate = 0.05;
```
zloery, line 14: This seems like a useful configuration option, why remove it?

zloery (commented):
Few comments on configuration but generally LGTM, thank you for taking this on!

src/main/java/bio/terra/common/tracing/TracingConfig.java
```diff
@@ -1,151 +1,36 @@
...
   @Bean
-  public FilterRegistrationBean<OcHttpServletFilter> tracingServletFilter() {
-    // Sam only supports B3 format headers, so prefer those everywhere to the default TraceContext
-    // format.
-    servletContext.setAttribute(
-        OcHttpServletFilter.OC_TRACE_PROPAGATOR, Tracing.getPropagationComponent().getB3Format());
-    FilterRegistrationBean<OcHttpServletFilter> registration =
-        new FilterRegistrationBean<>(new OcHttpServletFilter());
-    registration.setUrlPatterns(tracingProperties.getUrlPatterns());
-    return registration;
```
zloery, line 80: I think we previously relied on this servlet filter to set up the context for incoming requests (e.g. in `RequestAttributeInterceptor`). Is that automatic now?

build.gradle
```diff
@@ -14,7 +14,7 @@ plugins {
     // removed this plugin and filled in specific version numbers. I also removed all
     // version numbers with `+` in them. Was:
     //     id 'io.spring.dependency-management' version '1.0.10.RELEASE'
```
pshapiro4broad, line 16: I wonder if this is still true. It would be better (and more consistent with other terra projects) to use the dependency management plugin here.

settings.gradle
pshapiro4broad: I agree, although the version should be updated by the bumper, not as a manual commit.

To update the minor version you need to add a `#minor` tag to the PR commit message. https://github.com/DataBiosphere/github-actions/tree/master/actions/bumper#normal-flow

src/main/java/bio/terra/common/iam/SamUserFactory.java
```diff
@@ -28,16 +29,16 @@
...
+        ot -> apiClientBuilder.addInterceptor(new OkHttpClientTracingInterceptor(ot)));
```
pshapiro4broad: Personally I think `null` is fine, but if you want to use `Optional`, you may as well use `map()` too
```suggestion
    openTelemetry.map(OkHttpClientTracingInterceptor::new)
            .ifPresent(apiClientBuilder::addInterceptor);
```

src/main/java/bio/terra/common/iam/SamUserFactory.java
```diff
@@ -28,16 +29,16 @@
...
-          .newBuilder()
-          .addInterceptor(new OkHttpClientTracingInterceptor(Tracing.getTracer()))
-          .build();
+  final OkHttpClient httpClient;
 
-  @Autowired
-  public SamUserFactory(BearerTokenFactory bearerTokenFactory) {
+  @Autowired(required = false)
+  public SamUserFactory(
+      BearerTokenFactory bearerTokenFactory, Optional<OpenTelemetry> openTelemetry) {
```
pshapiro4broad: Similar to declaring a constructor arg `@Nullable`, declaring it as `Optional` should also set `required = false` so you don't need to change the annotation. If you leave `required = false`, that makes `BearerTokenFactory` nullable too which I don't think you want to do.
```suggestion
  @Autowired
  public SamUserFactory(
      BearerTokenFactory bearerTokenFactory, Optional<OpenTelemetry> openTelemetry) {
```

src/test/java/bio/terra/common/stairway/TestMetricExporter.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Collection;
+import java.util.Collections;
+
+/** A test metric exporter that stores the last metrics it received. */
+public class TestMetricExporter implements MetricExporter {
+  private Collection<MetricData> lastMetrics;
+
+  @Override
+  public synchronized CompletableResultCode export(Collection<MetricData> metrics) {
+    lastMetrics = Collections.unmodifiableCollection(new ArrayList<>(metrics));
```
pshapiro4broad: `List.copyOf()` is a more direct way to create a read-only list that's a copy of an existing colleciton
```suggestion
    lastMetrics = List.copyOf(metrics);
```

src/main/java/bio/terra/common/stairway/MetricsHelper.java
```diff
@@ -1,172 +1,118 @@
...
+  public static final Aggregation COUNT_AGGREGATION = Aggregation.sum();
+
+  /** Gauge for flight latency in milliseconds. */
+  private final LongHistogram FLIGHT_LATENCY;
+  /** Counter for number of errors from stairway flights. */
+  private final LongCounter FLIGHT_ERROR_COUNT;
+  /** Gauge for step latency in milliseconds. */
+  private final LongHistogram STEP_LATENCY;
+  /** Counter for number of errors from stairway steps. */
+  private final LongCounter STEP_ERROR_COUNT;
```
pshapiro4broad: All caps are used for static final members only. I would expect these names to be camel case instead.

src/test/java/bio/terra/common/stairway/MonitoringHookTest.java
```diff
@@ -20,31 +20,62 @@
...
+    return OpenTelemetrySdk.builder()
+        .setPropagators(propagators)
+        .setMeterProvider(sdkMeterProviderBuilder.build())
+        .build();
+  }
 
   @Test
   public void spansConnected() throws Exception {
     List<SpanContext> contextRecord = new ArrayList<>();
     RecordContextStep.setRecord(contextRecord);
```
pshapiro4broad, line 74: I know this is old code, but a setter here is unnecessary as an inner class's `private` fields aren't private with respect to its containing class. I would remove the setter and assign the field directly.
```suggestion
    RecordContextStep.contextRecord = contextRecord; 
```

It might be clearer to move `RecordContextStep.contextRecord` so it's a field on `MonitoringHookTest`. ~Once that's done it no longer needs to be `static` anymore either.~ Actually it has to be static somewhere because steps and flights must be static classes.

pshapiro4broad (approved):
The bulk of this was telemetry changes which I know nothing about and can't speak to. It looks like it was a lot of work though -- thanks @dvoet !

src/main/java/bio/terra/common/tracing/TracingConfig.java
dvoet: This reads it directly from config, no need to have it in TracingProperties too

src/main/java/bio/terra/common/tracing/TracingProperties.java
dvoet: it is configured in open telemetry directly
https://github.com/DataBiosphere/terra-java-project-template/pull/21/files#diff-4dab07009b19ea187f607a407b00997345505227a852e7115e4dcd086f27a388R89

src/main/java/bio/terra/common/tracing/TracingConfig.java
dvoet: `RequestAttributeInterceptor` is still a thing for terra specific values. The other attributes are set by open telemetry

settings.gradle
dvoet: hopefully I will remember

zloery (approved):

cahrens:
> > Can you also provide a description of what testing you have done? For instance, do the unit tests provide enough coverage that you are fairly confident everything is working? Did you test this new version with any of our existing services?
> 
> [DataBiosphere/terra-java-project-template#21](https://github.com/DataBiosphere/terra-java-project-template/pull/21)

It looks like it couldn't find the new version.

PR #122, opened 2023-11-21
[PF-2977] Azure Service Bus support for stairway

v-hsompalli:
1) Updated common-lib build file to use the latest version of stairway which supports Azure Service Bus(0.0.79)
2) Updated properties file to include Azure Service Bus related properties, **azureQueueEnabled** property should be true to use azure service bus queue as Work Queue. If it set to **false** so that gcp would be the Work Queue.
3) setupWorkQueue()  method from StairwayComponent is updated to return QueueInterface ,  so it will suport both GCP and Azure.


giventocode (approved):
LGTM

build.gradle
```diff
@@ -81,6 +81,7 @@ dependencies {
     // Terra libraries
     implementation group: 'org.broadinstitute.dsde.workbench', name: 'sam-client_2.13', version: '0.1-f554115'
     api group: 'bio.terra', name: 'stairway-gcp', version: '0.0.76-SNAPSHOT'
+    implementation group: 'bio.terra', name: 'stairway-azure', version: '0.0.79-SNAPSHOT'
```
dvoet: stairwayy-gcp and stairway-azure should use the same version, define it as a constant

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -54,7 +57,7 @@ public StairwayComponent(
    * clusterNameSuffix is present, then we create the queue. It is an error to try to run in
    * Kubernetes without a work queue.
    */
-  private GcpPubSubQueue setupWorkQueue() {
+  private QueueInterface setupWorkQueue() {
```
dvoet: rename to setupGcpWorkQueue?

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -91,6 +94,21 @@ private GcpPubSubQueue setupWorkQueue() {
     }
   }
 
+
+  /**
+   * Set up the Stairway Azure work queue.
+   */
+  public QueueInterface setupAzureWorkQueue() {
```
dvoet:
```suggestion
  private QueueInterface setupAzureWorkQueue() {
```

src/main/java/bio/terra/common/stairway/StairwayComponent.java
```diff
@@ -113,7 +131,13 @@ public void initialize(DataSource dataSource, Object context, List<StairwayHook>
    * @param initializeBuilder collection of Stairway initialization parameters
    */
   public void initialize(StairwayOptionsBuilder initializeBuilder) {
-    QueueInterface queue = (kubeProperties.isInKubernetes()) ? setupWorkQueue() : null;
+    QueueInterface queue;
+    //Using Azure WorkQueue if azureQueueEnabled set to true
+    if(stairwayProperties.isAzureQueueEnabled()) {
+      queue = setupAzureWorkQueue();
```
dvoet, line 132: why does gcp care about kubeProperties.isInKubernetes but azure does not?

v-hsompalli: This is existing code and Im not sure what is the reason behind it.

dvoet: I am suggesting you find out the reason. It is weird to me that gcp would care about k8s but azure would not.

src/main/java/bio/terra/common/stairway/StairwayComponent.java
dvoet: if this is used in tests, make this package private and use the `VisibleForTesting` annotation

dvoet (approved):

giventocode (approved):

giventocode (approved):

dvoet (approved):

dvoet (approved):

giventocode (approved):

rtitle (approved):
Looks good to me

PR #124, opened 2023-12-01
WOR-1383 default excluded urls

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-1383

### DIFF
--- a/src/main/java/bio/terra/common/opentelemetry/OpenTelemetryConfig.java
+++ b/src/main/java/bio/terra/common/opentelemetry/OpenTelemetryConfig.java
@@ -10,15 +10,14 @@ import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.sdk.trace.SdkTracerProvider;
 import io.opentelemetry.sdk.trace.SpanProcessor;
 import io.opentelemetry.sdk.trace.samplers.Sampler;
+import java.util.Optional;
+import java.util.Set;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Primary;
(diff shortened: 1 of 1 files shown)


cahrens (approved):

src/main/java/bio/terra/common/opentelemetry/OpenTelemetryConfig.java
```diff
@@ -52,7 +57,7 @@ public SdkTracerProvider terraTraceProvider(
     spanProcessors.stream().forEach(tracerProviderBuilder::addSpanProcessor);
     tracerProviderBuilder.setSampler(
         new ExcludingUrlSampler(
-            tracingProperties.excludedUrls(),
+            Optional.ofNullable(tracingProperties.excludedUrls()).orElse(DEFAULT_EXCLUDED_URLS),
```
pshapiro4broad, line 60: Using `Optional.of` creates a new object on the heap, you can avoid that by using `Objects.requireNonNullElse()`
```suggestion
            Objects.requireNonNullElse(tracingProperties.excludedUrls(), DEFAULT_EXCLUDED_URLS),
```

blakery (approved):

src/main/java/bio/terra/common/opentelemetry/OpenTelemetryConfig.java
```diff
@@ -21,6 +23,9 @@
 @EnableOpenTelemetry
 @EnableConfigurationProperties(value = {TracingProperties.class})
 public class OpenTelemetryConfig {
+
+  public static final Set<String> DEFAULT_EXCLUDED_URLS = Set.of("/status", "/version");
```
pshapiro4broad, line 27: Not a huge deal but I don't see why the first parameter to `ExcludingUrlSampler()` isn't `Collection<String>`. The code works the same if you pass in a `List` or `Set` and the collection is unlikely to be long enough for a set to give any benefit.

pshapiro4broad (approved):
I have comments, but it looks good to me overall.

PR #127, opened 2023-12-04
[PF-2978] High severity security issue - Upgrade logback-core

rtitle:
https://broadworkbench.atlassian.net/browse/PF-2978

[CVE-2023-6378](https://www.cve.org/cverecord?id=CVE-2023-6378) is fixed as of logback `1.4.14`: https://logback.qos.ch/news.html#1.3.14

However spring-boot pulls in logback versions which are vulnerable:
* `spring-boot:3.1.2` pulls in `logback-core:1.4.8`
* `spring-boot:3.2.0` pulls in `logback-core:1.4.11`

See also this spring-boot issue from today: https://github.com/spring-projects/spring-boot/issues/38643

So I'm leaving spring-boot alone, and explicitly pulling in logback from TCL, which seems to work.

Downstream services will need to update their TCL dependency.


src/main/java/bio/terra/common/logging/LoggingUtils.java
```diff
@@ -119,8 +118,7 @@ protected static void initializeLogging(ConfigurableApplicationContext applicati
         // seems to work. But if we encounter future issues in human-readable logging control, this
         // is a reasonable place to look more closely.
         logbackLogger.detachAndStopAllAppenders();
-        new ContextInitializer(logbackLogger.getLoggerContext())
-            .configureByResource(ResourceUtils.getURL("classpath:logback.xml"));
+        new ContextInitializer(logbackLogger.getLoggerContext()).autoConfig();
```
rtitle, line 121: I think this is the same thing. The old method was deprecrated and now removed, and said:
```
    /**
     * This method is deprecated with no replacement
     * @param url
     * @throws JoranException
     */
    @Deprecated
    public void configureByResource(URL url) throws JoranException {
```

https://github.com/qos-ch/logback/blob/v_1.4.8/logback-classic/src/main/java/ch/qos/logback/classic/util/ContextInitializer.java#L55-L61

aherbst-broad (approved):

dvoet (approved):

PR #129, opened 2024-01-04
[IA-4756] remove spring.factories

LizBaldo:
The `bio.terra.common.tracing.disable.AutoConfigDisable` file was previously removed in https://github.com/DataBiosphere/terra-common-lib/pull/119 and so we should remove any configuration based on this file.

This is a follow up from the work being done to upgrade the terra azure rely listener repo to the latest TCL https://broadworkbench.atlassian.net/browse/IA-4756


dvoet (approved):

cpate4 (approved):
Based on comments in slack.

PR #130, opened 2024-01-04
Upgrade to Spring Boot 3.2.1; remove explicit logback dependency

rtitle:
Basically un-does https://github.com/DataBiosphere/terra-common-lib/pull/127/, now that Spring Boot has upgraded to a patched `logback-core`.


build.gradle
```diff
@@ -14,7 +14,7 @@ plugins {
     // removed this plugin and filled in specific version numbers. I also removed all
     // version numbers with `+` in them. Was:
     //     id 'io.spring.dependency-management' version '1.0.10.RELEASE'
-    id 'org.springframework.boot' version '3.1.2'
+    id 'org.springframework.boot' version '3.2.1'
```
rtitle, line 17: Any downside in moving TCL to `3.2`? I didn't see anything that would drastically affect us.

https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.2-Release-Notes

rtitle:
Hm tests failed, maybe there is some issue with the new dependencies. Will dig more later..

cahrens:
@rtitle, we are now at version 3.2.3 (my dependabot ignore was incorrect), so I'm trying this again in https://github.com/DataBiosphere/terra-common-lib/pull/156.

PR #131, opened 2024-02-08
WOR-1510 upgrade otel override HttpServerMetrics

dvoet:
https://broadworkbench.atlassian.net/browse/WOR-1510


src/main/java/bio/terra/common/opentelemetry/HttpMetricsAdvice.java
```diff
@@ -0,0 +1,81 @@
...
+                if (d < 2) {
+                  return d + 0.1;
+                }
+                if (d < 10) {
+                  return d + 1;
+                }
+                if (d < 60) {
+                  return d + 5;
+                }
+                return d + 10;
```
tlangs, line 43: Since this is Java 17, this could be a switch with guards.

dvoet: copilot wrote this code :)

tlangs: 🤖 🚫

dvoet: I think guards is a preview feature of java 17 and we don't have that enabled

okotsopoulos (approved):
Looks good as far as I can tell.

Does a service using OTEL via TCL need to do anything extra to get these changes?  Is there anywhere I can see these updated metrics in action?

src/main/java/bio/terra/common/opentelemetry/HttpMetricsAdvice.java
tlangs: ah, darn

tlangs (approved):

dvoet:
> Does a service using OTEL via TCL need to do anything extra to get these changes?

yes, they need to upgrade their TCL version, I will start doing that tomorrow

PR #132, opened 2024-02-09
[WOR-1502] Add dependabot configuration.

cahrens:
Added a configuration based off of [WSM's configuration](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/.github/dependabot.yml).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,26 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    open-pull-requests-limit: 20
+    groups:
+      minor-patch-dependencies:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
+    schedule:
+      interval: "monthly"
+      time: "06:00"
+      timezone: "America/New_York"
+    target-branch: "main"
+    reviewers:
+      - "@DataBiosphere/platform-foundation-codeowners"
+    labels:
+      - "dependency"
+      - "gradle"
+    commit-message:
+      prefix: "[PF-2983]"
+    ignore:
+      - dependency-name: "org.springframework.boot:spring-boot-gradle-plugin"


.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+    open-pull-requests-limit: 20
+    groups:
+      minor-patch-dependencies:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
+    schedule:
+      interval: "monthly"
```
cahrens, line 14: We wanted every 2 weeks, but that is not a supported option (options are daily, weekly, and monthly). I decided to go with monthly over weekly, but with a higher open pull request limit (20).

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    open-pull-requests-limit: 20
+    groups:
+      minor-patch-dependencies:
```
cahrens, line 7: Group minor and patch updates together. Leave major updates as individual PRs since they are more likely to involve a breaking change. This way of grouping PRs has been working well for the Workspaces team.

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+        update-types:
+          - "minor"
+          - "patch"
+    schedule:
+      interval: "monthly"
+      time: "06:00"
+      timezone: "America/New_York"
+    target-branch: "main"
+    reviewers:
+      - "@DataBiosphere/platform-foundation-codeowners"
```
cahrens, line 19: https://github.com/orgs/DataBiosphere/teams/platform-foundation-codeowners

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+    labels:
+      - "dependency"
+      - "gradle"
+    commit-message:
+      prefix: "[PF-2983]"
```
cahrens, line 24: Long-lived ticket, https://broadworkbench.atlassian.net/browse/PF-2983

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+    ignore:
+      - dependency-name: "org.springframework.boot:spring-boot-gradle-plugin"
```
cahrens, line 26: I added an ignore based on the open PR, https://github.com/DataBiosphere/terra-common-lib/pull/130

.github/dependabot.yml
rushtong: Something we like to do is add labels which make the PR list a little more visual when viewing https://github.com/pulls/review-requested

Screenshot 2024-02-09 at 11 57 08 AM

There are other times when we manually do library updates so we reserve that prefix for dev-initiated PRs and a slightly different one for dependabot-initiated PRs.

rushtong (approved):
Everything looks reasonable to me 👍🏽

.github/dependabot.yml
cahrens: @rushtong does the automatic linking to the JIRA ticket work with "-dependabot" appended?

.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    open-pull-requests-limit: 20
```
cjllanwarne, line 5: Do we know how long the initial backlog is? Will 20 per month keep up with the pace of new arrivals? 

Do you know how the top 20 updates will be sorted and chosen? (date created first? major first? ( / minor first?) / alphabetically first?)

cahrens: I do not know how long the backlog is, but we can always adjust this if necessary. With the grouping of minor/patch, it's possible that 20 will be plenty.

I don't see anything in the documentation about how it decides the ordering (https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#frequency-of-dependabot-pull-requests). Let me know if you find something!

.github/dependabot.yml
rushtong: It does link in our board.

rtitle (approved):
I don't have much experience configuring dependabot, but this looks good to me! Thanks for doing this @cahrens

.github/dependabot.yml
cahrens: OK, we can always tweak the prefix if people think it helps (I've always found that the titles and labels make it pretty clear).

PR #133, opened 2024-02-12
[WOR-1502] Update "main" branch

cahrens:
There is a main branch in terra-common-lib, but it only has 1 commit… perhaps as part of our repo standardization we can also standardize on this?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,7 +14,7 @@ updates:
       interval: "monthly"
       time: "06:00"
       timezone: "America/New_York"
-    target-branch: "main"
+    target-branch: "develop"
     reviewers:
       - "@DataBiosphere/platform-foundation-codeowners"
     labels:


pshapiro4broad:
> There is a main branch in terra-common-lib, but it only has 1 commit… perhaps as part of our repo standardization we can also standardize on this?

Not sure if we need to standardize on the name of the `main` branch in repos, but in this repo I think we should delete the `main` branch to avoid further confusion. I see a number of other stale branches that could be deleted too.

pshapiro4broad (approved):

cahrens:
> Not sure if we need to standardize on the name of the `main` branch in repos, but in this repo I think we should delete the `main` branch to avoid further confusion. I see a number of other stale branches that could be deleted too.

I would love for us to have a standard main branch name. I have this confusion all the time when working on repos-- is it main, develop, or dev? I was assuming here that it's main because that is what the other Java repos use, but surprisingly, no.

cjllanwarne (approved):

PR #158, opened 2024-02-27
[PF-2988] Remove flagsmith

cahrens:
Verily no longer uses this repo, and DSP isn't using Flagsmith.


rtitle (approved):

cahrens:
I searched and did not find any terra helmfile references to flagsmith. 

We have some code in WSM that we will need to excise when we upgrade the TCL version there. https://github.com/search?q=terra.common.flagsmith&type=code

Theoretically this PR should have a major version bump, but we just did a major version bump yesterday (because of the large number of 3rd party updates going in) and downstream services haven't adopted the new version yet.

pshapiro4broad (approved):

PR #159, opened 2024-02-29
ID-1105 Remediate CVE-2024-26308 (Apache Commons Compress) 

tlangs:
Forcing Apache Commons Compress to version 1.26.0 to remediate https://nvd.nist.gov/vuln/detail/CVE-2024-26308

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -131,6 +131,8 @@ dependencies {
         implementation('org.json:json:20240205')
         implementation('org.bitbucket.b_c:jose4j:0.9.5')
         implementation('io.grpc:grpc-xds:1.61.1')
+        // commons-compress comes from io.kubernetes:client-java, which doesn't yet have a version
+        // that pulls in the non-vulnerable version of commons-compress.
         implementation('org.apache.commons:commons-compress:1.26.0')
     }
 }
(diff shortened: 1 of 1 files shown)


Shakespeared (approved):

build.gradle
```diff
@@ -131,6 +131,7 @@ dependencies {
         implementation('org.json:json:20240205')
         implementation('org.bitbucket.b_c:jose4j:0.9.5')
         implementation('io.grpc:grpc-xds:1.61.1')
+        implementation('org.apache.commons:commons-compress:1.26.0')
```
okotsopoulos, line 136: If this is coming in via `io.kubernetes:client-java`, it would be nice to have a comment to that effect.  I know we need to constrain because the k8s client hasn't fixed the vulnerability on its end, but hopefully it will one day.  We've found that keeping constraints around indefinitely can increase the maintenance burden.

okotsopoulos (approved):
Looks good but would love an extra comment in there.

build.gradle
tlangs: k8s-client doesn't yet have a version that pulls in the fixed dep. I'm also nervous about updating the k8s-client. Perhaps that's more a condemnation of TCL as a concept. 

I'll add that comment!

PR #164, opened 2024-03-19
[WOR-1560] Derive opentelemetry versions from BOMs

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1560

When available, using a BOM (bill of materials) to synchronize related dependency versions makes it less likely that an incompatible upgrade is introduced.

Our OpenTelemetry dependencies can all be derived from [BOMs](https://github.com/open-telemetry/opentelemetry-java?tab=readme-ov-file#overview):
- [opentelemetry-bom](https://mvnrepository.com/artifact/io.opentelemetry/opentelemetry-bom) synchronizes stable `io.opentelemetry:*` dependencies
- [opentelemetry-bom-alpha](https://mvnrepository.com/artifact/io.opentelemetry/opentelemetry-bom-alpha) synchronizes unstable `io.opentelemetry:*` dependencies
- [opentelemetry-instrumentation-bom](https://mvnrepository.com/artifact/io.opentelemetry.instrumentation/opentelemetry-instrumentation-bom) synchronizes stable `io.opentelemetry.instrumentation:*` dependencies
- [opentelemetry-instrumentation-bom-alpha](https://mvnrepository.com/artifact/io.opentelemetry.instrumentation/opentelemetry-instrumentation-bom-alpha) synchronizes unstable `io.opentelemetry.instrumentation:*` dependencies

We pulled in dependencies from all of the above BOMs.  They now derive their versions from these BOMs.

While here...
- Expose `io.opentelemetry:opentelemetry-api` as an `api` dependency rather than `implementation`, so that services pulling in TCL won't have to define it themselves.
- Reformatted dependency definitions so that Dependabot will register them: I think that Dependabot may not pick up on those formatted strings of ours.

I also tried to version BOMs for latest dependencies, which required a slight adjustment to our redefinition of HttpMetricsAdvice (I used the latest version of [OTEL source code's HttpMetricsAdvice](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/release/v2.2.x/instrumentation-api/src/main/java/io/opentelemetry/instrumentation/api/semconv/http/HttpMetricsAdvice.java) as a reference).  **But**, [tests failed](https://scans.gradle.com/s/gjvj4cxmcaqss/tests/overview?outcome=FAILED) and I decided it was better to allow any upgrades to be handled separately from the move to BOMs for versioning.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -89,24 +89,26 @@ dependencies {
     implementation group: 'ch.qos.logback.contrib', name: 'logback-json-classic', version: '0.1.5'
     implementation group: 'ch.qos.logback.contrib', name: 'logback-jackson', version: '0.1.5'
 
-    // OpenTelemetry
-    var openTelemetryVersion = '1.34.1'
-    var openTelemetryInstrumentationVersion = '2.0.0'
-    implementation "io.opentelemetry:opentelemetry-api:${openTelemetryVersion}"
-    implementation "io.opentelemetry:opentelemetry-sdk:${openTelemetryVersion}"
-    implementation "io.opentelemetry:opentelemetry-sdk-metrics:${openTelemetryVersion}"
-    implementation "io.opentelemetry:opentelemetry-exporter-logging:${openTelemetryVersion}"
-    implementation "io.opentelemetry.instrumentation:opentelemetry-spring-webmvc-6.0:${openTelemetryInstrumentationVersion}-alpha"
-    implementation "io.opentelemetry.instrumentation:opentelemetry-instrumentation-api:${openTelemetryInstrumentationVersion}"
-    implementation "io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:${openTelemetryInstrumentationVersion}"
-    implementation "io.opentelemetry.instrumentation:opentelemetry-spring-boot:${openTelemetryInstrumentationVersion}-alpha"
-    implementation "io.opentelemetry:opentelemetry-exporter-prometheus:${openTelemetryVersion}-alpha"
-    implementation "io.opentelemetry:opentelemetry-extension-incubator:${openTelemetryVersion}-alpha"
+    // OpenTelemetry BOMs
+    implementation platform('io.opentelemetry:opentelemetry-bom:1.36.0')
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -89,24 +89,26 @@ dependencies {
...
-    implementation "io.opentelemetry:opentelemetry-exporter-logging:${openTelemetryVersion}"
-    implementation "io.opentelemetry.instrumentation:opentelemetry-spring-webmvc-6.0:${openTelemetryInstrumentationVersion}-alpha"
-    implementation "io.opentelemetry.instrumentation:opentelemetry-instrumentation-api:${openTelemetryInstrumentationVersion}"
-    implementation "io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:${openTelemetryInstrumentationVersion}"
-    implementation "io.opentelemetry.instrumentation:opentelemetry-spring-boot:${openTelemetryInstrumentationVersion}-alpha"
-    implementation "io.opentelemetry:opentelemetry-exporter-prometheus:${openTelemetryVersion}-alpha"
-    implementation "io.opentelemetry:opentelemetry-extension-incubator:${openTelemetryVersion}-alpha"
+    // OpenTelemetry BOMs
+    implementation platform('io.opentelemetry:opentelemetry-bom:1.34.1')
+    implementation platform('io.opentelemetry:opentelemetry-bom-alpha:1.34.1-alpha')
```
cahrens, line 94: Why do we need the alpha one?

cahrens: Ah nevermind, I read your PR summary.

cahrens (approved):

build.gradle
okotsopoulos: To answer your question more concretely, here's what we previously pulled in that we now get from this Alpha BOM:

```
    implementation "io.opentelemetry:opentelemetry-exporter-prometheus:${openTelemetryVersion}-alpha"
    implementation "io.opentelemetry:opentelemetry-extension-incubator:${openTelemetryVersion}-alpha"
```

mmorgantaylor (approved):
thank you for doing this!

PR #167, opened 2024-03-29
[PF-3000] Version io.kubernetes:client-java at 20.0.1-legacy instead of 20.0.1

okotsopoulos:
https://broadworkbench.atlassian.net/browse/PF-3000

**Background**

`terra-common-lib` 1.1.0 upgraded its k8s client from 16.0.0 to 20.0.1 to address a number of security vulnerabilities (note that versions 20.0.0 onward without `-legacy` suffix contained breaking changes): https://github.com/DataBiosphere/terra-common-lib/pull/146

This upgrade appears to have broken [KubePodListener](https://github.com/DataBiosphere/terra-common-lib/blob/e7206005a1c33b4c20065c23bdf188ce12953468/src/main/java/bio/terra/common/kubernetes/KubePodListener.java), which is used by services using [Stairway with pubsub queues](https://broadinstitute.slack.com/archives/C02LMB1PQ6R/p1706809529949099).

The listener exhausts its retries because it can’t [deserialize the Watch response](https://github.com/DataBiosphere/terra-common-lib/blob/e7206005a1c33b4c20065c23bdf188ce12953468/src/main/java/bio/terra/common/kubernetes/KubePodListener.java#L101), and so it never gets to recover flights from deleted pods ([Stackdriver log](https://cloudlogging.app.goo.gl/9YXifWnQzPwRPYa76)).

This is because v20.0.1 includes validation on deserialization that was not previously there, nor present in legacy versions:

v20.0.1 -  https://github.com/kubernetes-client/java/blob/automated-release-20.0.1/kubernetes/src/main/java/io/kubernetes/client/openapi/models/V1NamespaceSpec.java

Versus v20.0.1-legacy - https://github.com/kubernetes-client/java/blob/automated-release-20.0.1-legacy/kubernetes/src/main/java/io/kubernetes/client/openapi/models/V1NamespaceSpec.java

**Changes**

Version the k8s client at 20.0.1-legacy instead of 20.0.1, in part achieved via a revert of the past upgrade.

By versioning the k8s client at 20.0.1-legacy, we address the same bouncycastle vulnerabilities as the prior non-legacy upgrade without the breaking changes.

Sourceclear has helped me feel confident in the state of this repo's dependency vulnerabilities, and that I've addressed the vulnerabilities I expect without bringing in new ones.  I added Gradle's srcclr plugin and documented how to run scans locally or off of a published feature branch via Github Actions, matching the mechanism and settings that AppSec uses when running these scans.

I've also instructed Dependabot to ignore k8s client upgrades.

**Instructions for Consumers of TCL**

If you haven't already upgraded to TCL 1.1.0+, upgrade TCL as normal.  Do not upgrade k8s client to a non-legacy version.
Otherwise, if your service upgraded k8s client to 20.0.0+ (non-legacy), you will need to revert your upgrade in favor of the legacy version.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,3 +22,6 @@ updates:
       - "gradle"
     commit-message:
       prefix: "[PF-2983]"
+    ignore:
+      # PF-3000: k8s client versions 20.0.0 onward without '-legacy' suffix include breaking changes
+      - dependency-name: "io.kubernetes:client-java"
--- a/build.gradle
+++ b/build.gradle
@@ -58,7 +58,8 @@ dependencies {
     annotationProcessor group: 'org.springframework.boot', name: 'spring-boot-configuration-processor', version: '3.2.3'
 
     // Misc. Services
-    implementation group: 'io.kubernetes', name: 'client-java', version: '20.0.1-legacy'
+    // PF-3000: k8s client versions 20.0.0 onward without '-legacy' suffix include breaking changes
+    implementation group: 'io.kubernetes', name: 'client-java', version: '20.0.1-legacy' // keep legacy suffix
     implementation group: 'io.prometheus', name: 'simpleclient_httpserver', version: '0.16.0'
     implementation group: 'jakarta.ws.rs', name: 'jakarta.ws.rs-api', version: '3.1.0'
 
--- a/build.gradle
+++ b/build.gradle
@@ -15,6 +15,7 @@ plugins {
     //     id 'io.spring.dependency-management' version '1.0.10.RELEASE'
     id 'org.springframework.boot' version '3.2.3'
     id 'ru.vyarus.quality' version '5.0.0'
+    id 'com.srcclr.gradle' version '3.1.12'
 }
 
 group = gradle.projectGroup
@@ -155,6 +156,7 @@ apply from: "$gradleIncDir/jacoco.gradle"
 apply from: "$gradleIncDir/javadoc.gradle"
 apply from: "$gradleIncDir/publishing.gradle"
 apply from: "$gradleIncDir/quality.gradle"
+apply from: "$gradleIncDir/srcclr.gradle"
 apply from: "$gradleIncDir/sonarqube.gradle"
 apply from: "$gradleIncDir/spotless.gradle"
 apply from: "$gradleIncDir/testing.gradle"
--- a/build.gradle
+++ b/build.gradle
@@ -131,11 +131,7 @@ dependencies {
     constraints {
         implementation('org.scala-lang:scala-library:2.13.13')
         implementation('org.json:json:20240205')
-        implementation('org.bitbucket.b_c:jose4j:0.9.5')
         implementation('io.grpc:grpc-xds:1.62.2')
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -131,6 +131,7 @@ dependencies {
     constraints {
         implementation('org.scala-lang:scala-library:2.13.13')
         implementation('org.json:json:20240205')
+        implementation('org.bitbucket.b_c:jose4j:0.9.5')
```
okotsopoulos: This was part of the revert, but it's not needed: k8s client 20.0.1-legacy addresses this vulnerability.  I'll remove it.

build.gradle
```diff
@@ -132,9 +132,6 @@ dependencies {
         implementation('org.scala-lang:scala-library:2.13.13')
         implementation('org.json:json:20240205')
         implementation('io.grpc:grpc-xds:1.62.2')
-        // commons-compress comes from io.kubernetes:client-java, which doesn't yet have a version
-        // that pulls in the non-vulnerable version of commons-compress.
-        implementation('org.apache.commons:commons-compress:1.26.0')
```
okotsopoulos, line 137: No longer the case: k8s client `20.0.1-legacy` pulls in this version.

.github/dependabot.yml
```diff
@@ -22,3 +22,6 @@ updates:
       - "gradle"
     commit-message:
       prefix: "[PF-2983]"
+    ignore:
+      # PF-3000: k8s client versions 20.0.0 onward without '-legacy' suffix include breaking changes
+      - dependency-name: "io.kubernetes:client-java"
```
mmorgantaylor, line 27: is it safe to ignore this dependency long-term? do we need to add a ticket to address the breaking changes and remove this ignore statement?

build.gradle
```diff
@@ -57,7 +58,7 @@ dependencies {
     annotationProcessor group: 'org.springframework.boot', name: 'spring-boot-configuration-processor', version: '3.2.3'
 
     // Misc. Services
-    implementation group: 'io.kubernetes', name: 'client-java', version: '20.0.1'
+    implementation group: 'io.kubernetes', name: 'client-java', version: '20.0.1-legacy'
```
cahrens: Probably worth adding a comment here about sticking with the "legacy" suffix.

cahrens (approved):
Thanks!

build.gradle
okotsopoulos: Good call.

mmorgantaylor (approved):
looks reasonable to me. happy to test with TSPS as soon as this is merged.

.github/dependabot.yml
okotsopoulos: I kind of overloaded the https://broadworkbench.atlassian.net/browse/PF-3000 ticket by attaching it to this change.  But I can update it to mention that if we safely get off the legacy version, we might be able to remove this ignore.

However, we may also want to keep the ignore in place, unless we can figure out how to tell Dependabot to either only update with `-legacy` versions, or ignore `-legacy` versions.  Right now as far as I can see there's no way to do that.

PR #173, opened 2024-03-29
[PF-2953] Use Spring dependency manager

okotsopoulos:
Previously, this was disabled because it generated invalid POM and broke publishing to mavenLocal, which is useful for testing.

This is no longer the case: I verified that when bringing Spring dependency management back in at the latest release I could still publish to mavenLocal and pull in the resulting artifact into another service.

As a result, many places where we previously versioned dependencies can now derive their versions from the dependency manager.

In many places, we were defining dependencies redundantly: we're pulling in a number of Spring starters which expose more transitive dependencies for our usage than we realized.

I did not upgrade any versions to keep the scope narrow: I'll leave that to Dependabot.  But by delegating dependency management to Spring, we'll have fewer dependencies to upgrade directly and stand a better chance of keeping them compatible with one another.


build.gradle
```diff
@@ -86,13 +81,11 @@ dependencies {
     implementation "bio.terra:stairway-azure:${stairwayVersion}"
 
     // Logging
-    implementation group: 'org.slf4j', name: 'slf4j-api', version: '2.0.12'
```
cahrens, line 89: Do you recall why this is no longer needed?

cahrens (approved):

build.gradle
okotsopoulos: We get it as a transitive implementation dependency via `spring-boot-starter-web`.  And the 2.0.12 version comes from the Spring dependency manager:

https://github.com/spring-projects/spring-boot/blob/v3.2.3/spring-boot-project/spring-boot-dependencies/build.gradle#L1474-L1490

build.gradle
```diff
@@ -40,28 +40,23 @@ repositories {
 }
 
 dependencies {
-    // Common utils
-    implementation group: 'org.apache.commons', name: 'commons-dbcp2', version: '2.11.0'
-    implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.14.0'
-    implementation group: 'org.apache.commons', name: 'commons-pool2', version: '2.12.0'
+    // Common utils (unversioned are versioned by Spring dependency manager)
```
pshapiro4broad: It looks like there are now many cases where versions are provided by spring. Do you think we need a comment on each case? If a new dependency is added without a comment, do you think it will be confusing for future readers?

I was thinking that, in general, we should have good tools to help us determine where a version is coming from, so if the need arises we can figure it out, rather than trying to comment each instance individually. But maybe you find this approach better?

pshapiro4broad (approved):
Looks good to me, I'm glad you were able to get this working!

build.gradle
okotsopoulos: Good question.  The comments throughout felt like overkill as I was writing them and I agree this could get out of date.  But for this repo, this was a pretty big shift in approach, and I have found that analyzing dependencies is not yet a commonly-held competency (I have intentionally built this up for myself over the past few months).

I'll replace the comments throughout with a top-level comment though.

PR #174, opened 2024-04-01
fix KubePodListener and remove legacy k8s api

dvoet:
https://broadworkbench.atlassian.net/browse/PF-3000

The problem was that `KubePodListenter` listens for _pods_ but the type the code used was `V1Namespace`. The new java client adds validation and, as one might expected, a response for pods has unexpected fields for a namespace. The correction is to switch from `V1Namespace` to `V1Pod`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,6 +22,3 @@ updates:
       - "gradle"
     commit-message:
       prefix: "[PF-2983]"
-    ignore:
-      # PF-3000: k8s client versions 20.0.0 onward without '-legacy' suffix include breaking changes
-      - dependency-name: "io.kubernetes:client-java"
--- a/build.gradle
+++ b/build.gradle
@@ -58,8 +58,7 @@ dependencies {
     annotationProcessor group: 'org.springframework.boot', name: 'spring-boot-configuration-processor', version: '3.2.3'
 
     // Misc. Services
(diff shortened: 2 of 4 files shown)


.github/dependabot.yml
```diff
@@ -22,6 +22,3 @@ updates:
       - "gradle"
     commit-message:
       prefix: "[PF-2983]"
-    ignore:
-      # PF-3000: k8s client versions 20.0.0 onward without '-legacy' suffix include breaking changes
-      - dependency-name: "io.kubernetes:client-java"
```
okotsopoulos, line 27: I think this is okay to take out in this PR, but flagging that we might ultimately add it back in if k8s client keeps on publishing legacy and non-legacy versions.  If they do, Dependabot will pick up on the last published version regardless of its legacy status.  I don't think there's a way to tell Dependabot to allow or exclude versions adhering to a naming convention (outside of the semver major, minor, patch designations).

mmorgantaylor (approved):
thanks!

okotsopoulos (approved):
Thanks a lot for digging into this, and verifying in a BEE (sharing those custom [TSPS logs](https://cloudlogging.app.goo.gl/ea6RLUbjskzkvPNM9) here for posterity).

Is there any reasonable way to get some automated test coverage or type-safety protections before runtime?  I haven't seen k8s client's Watch used anywhere else here or in downstream services, but it's concerning that this type mismatch was latent for years.

PR #177, opened 2024-04-01
[PF-2983] Remove unused protobuf dependency

okotsopoulos:
I found no reference to this implementation dependency in TCL.  Because it's not declared as an api dependency, I think we're safe to remove it.

Better to remove rather than spend effort updating it in https://github.com/DataBiosphere/terra-common-lib/pull/171 (or needing to chase future vulnerabilities within an unneeded dependency).


tlangs (approved):

nmalfroy (approved):
thanks!  I much prefer tying the protobuf version to what gcloud needs

PR #192, opened 2024-07-16
DR-3481 disable otel metrics exporter

dvoet:
https://broadworkbench.atlassian.net/browse/DR-3481


okotsopoulos (approved):
Thanks!  I think this needs a ticket but otherwise looks good.

Prior to this change, TDR was seeing an error log like this every 60 seconds:
```
2024-07-15 10:46:49,246 ERROR [OkHttp http://localhost:4318/...] i.o.e.internal.http.HttpExporter: Failed to export metrics. The request could not be executed. Full error message: Failed to connect to localhost/[0:0:0:0:0:0:0:1]:4318
java.net.ConnectException: Failed to connect to localhost/[0:0:0:0:0:0:0:1]:4318
```

Following guidelines for [local testing](https://github.com/DataBiosphere/terra-common-lib?tab=readme-ov-file#local-testing), I ran a local TDR using this PR's feature branch and no longer see those error logs.

pshapiro4broad (approved):
LGTM

PR #211, opened 2024-12-11
CORE-69: Spring Boot 3.4.0 and otel cleanup

davidangb:
* Upgrade to Spring Boot 3.4.0
* Align OpenTelemetry dependencies to the Spring Boot version and to each other
* Remove usages of `@SpyBean`, which is now deprecated
* bonus: update `sam-client` to latest

This PR supersedes #210

References: 
https://docs.spring.io/spring-boot/appendix/dependency-versions/coordinates.html
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.4-Release-Notes#deprecation-of-mockbean-and-spybean

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,11 +20,6 @@ project.ext {
     isCiServer = System.getenv().containsKey("CI")
 }
 
-// Spring Boot 3.4.0 pulls in opentelemetry-bom 1.43.0.
-// When upgrading Spring Boot, re-check these versions.
-ext['opentelemetry.version'] = '1.43.0'
-ext['opentelemetry.instrumentation.version'] = '2.9.0' // 2.9.0 targets opentelemetry 1.43.0
-
 // If true, search local repository (~/.m2/repository/) first for dependencies.
 def useMavenLocal = false
 repositories {
@@ -90,7 +85,10 @@ dependencies {
     implementation group: 'ch.qos.logback.contrib', name: 'logback-jackson', version: '0.1.5'
 
     // OpenTelemetry dependencies:
-    implementation platform("io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:${project.ext['opentelemetry.instrumentation.version']}-alpha")
+    // Spring Boot 3.4.0 pulls in opentelemetry-bom 1.43.0.
+    // We choose opentelemetry-instrumentation-bom-alpha:2.9.0-alpha because otel-instrumentation 2.9.0 targets otel 1.43.0.
+    // When upgrading Spring Boot, re-check these versions.
+    implementation platform("io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:2.9.0-alpha")
     // ... versioned by Spring Boot
     api 'io.opentelemetry:opentelemetry-api'
     implementation 'io.opentelemetry:opentelemetry-exporter-logging'
--- a/build.gradle
+++ b/build.gradle
@@ -79,7 +79,7 @@ dependencies {
     runtimeOnly group: 'org.postgresql', name: 'postgresql'
 
     // Terra libraries
-    implementation group: 'org.broadinstitute.dsde.workbench', name: 'sam-client_2.13', version: '0.1-0c4b377'
+    implementation group: 'org.broadinstitute.dsde.workbench', name: 'sam-client_2.13', version: 'v0.0.329'
     var stairwayVersion= '1.1.15-SNAPSHOT'
     api "bio.terra:stairway-gcp:${stairwayVersion}"
     implementation "bio.terra:stairway-azure:${stairwayVersion}"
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -51,6 +52,7 @@ dependencies {
 
     // Spring
     implementation group: 'org.springframework.retry', name: 'spring-retry'
+    implementation group: 'org.springframework.boot', name: 'spring-boot-autoconfigure'
     implementation group: 'org.springframework.boot', name: 'spring-boot-starter-data-jdbc'
     implementation group: 'org.springframework.boot', name: 'spring-boot-starter-web'
```
davidangb, line 52: `spring-boot-autoconfigure` moved up from line 108, where it was hidden in with otel dependencies

build.gradle
```diff
@@ -20,9 +20,10 @@ project.ext {
     isCiServer = System.getenv().containsKey("CI")
 }
 
-// Spring Boot 3.2.3 pulls in opentelemetry-bom 1.31.0.
+// Spring Boot 3.4.0 pulls in opentelemetry-bom 1.43.0.
 // We need >= 1.32.0 so that our HttpServerMetrics can use Meter.setExplicitBucketBoundariesAdvice:
-ext['opentelemetry.version'] = '1.42.1'
+ext['opentelemetry.version'] = '1.43.0'
+ext['opentelemetry.instrumentation.version'] = '2.9.0' // 2.9.0 targets opentelemetry 1.43.0
```
davidangb: this is actually a downgrade. We had been on `2.10.0` before this PR. I thought this was more appropriate, since `2.10.0` wants opentelemetry 1.44.1. See https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases.

If we decide to keep all the otel dependency versions aligned with Spring Boot, we should probably tell dependabot to ignore those updates.

build.gradle
```diff
@@ -87,25 +89,21 @@ dependencies {
...
     implementation 'io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations'
+    implementation 'io.opentelemetry.instrumentation:opentelemetry-instrumentation-api'
     implementation 'io.opentelemetry.instrumentation:opentelemetry-spring-boot-autoconfigure'
-    implementation 'io.opentelemetry:opentelemetry-exporter-prometheus'
+    implementation 'io.opentelemetry.semconv:opentelemetry-semconv'
     implementation 'io.opentelemetry:opentelemetry-api-incubator'
-    implementation 'io.opentelemetry:opentelemetry-sdk-extension-autoconfigure'
-    implementation 'io.opentelemetry:opentelemetry-exporter-otlp'
-    implementation 'org.springframework.boot:spring-boot-autoconfigure'
+    implementation 'io.opentelemetry:opentelemetry-exporter-prometheus'
```
davidangb, line 103: I reorganized this whole section. The diff is hard to read, so I'll try to summarize:

1. Moved the version number for `opentelemetry-instrumentation-bom-alpha` into an `ext` variable at line 26, so it is easily visible next to the main opentelemetry version.

2. removed a few dependencies which were redundant; these are pulled in transitively when needed:
```
implementation platform('io.opentelemetry:opentelemetry-bom-alpha:1.44.1-alpha')
implementation platform('io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom:2.10.0')
implementation 'io.opentelemetry:opentelemetry-sdk'
implementation 'io.opentelemetry:opentelemetry-sdk-metrics'
```

3. replaced `'io.opentelemetry.instrumentation:opentelemetry-spring-webmvc-6.0'` with `'io.opentelemetry.semconv:opentelemetry-semconv'`. The latter is smaller; we didn't really need the former.

4. grouped into two categories: those version-managed by Spring and those version-managed by us via `opentelemetry-instrumentation-bom-alpha`

5. within those groupings, alphabetized the libraries

build.gradle
```diff
@@ -77,7 +79,7 @@ dependencies {
     runtimeOnly group: 'org.postgresql', name: 'postgresql'
 
     // Terra libraries
-    implementation group: 'org.broadinstitute.dsde.workbench', name: 'sam-client_2.13', version: '0.1-0c4b377'
+    implementation group: 'org.broadinstitute.dsde.workbench', name: 'sam-client_2.13', version: 'v0.0.329'
```
davidangb, line 77: this isn't required by the Spring Boot upgrade, but while in this file I noticed that we were using an old, non-semantic-version of sam-client.

`sam-client:0.1-0c4b377` - Feb 14 2024
`sam-client:v0.0.329` - Dec 10 2024

davidangb:
@dvoet I'd especially love your opinion on the otel library-version changes here

calypsomatic (approved):

build.gradle
dvoet: do we need these lines anymore? I think they were there to specify a more advanced version than spring required, is that still true?

dvoet (approved):
looks good but one question

davidangb:
> do we need these lines anymore? I think they were there to specify a more advanced version than spring required, is that still true?

@dvoet I suspect we don't need them, but I was worried I missed some code somewhere that referenced that property. I will (try to) remove them.

PR #221, opened 2025-02-03
Update actions to current versions to fix build errors

pshapiro4broad:
Fix issue where builds are failing now that various github actions have been completely deprecated are are no longer supported. See https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/


davidangb (approved):

rjohanek (approved):
looks good!

PR #223, opened 2025-02-05
TSPS-404 Throw correct exception for Sam user not found

mmorgantaylor:
Teaspoons [uses](https://github.com/DataBiosphere/terra-scientific-pipelines-service/blob/main/service/src/main/java/bio/terra/pipelines/app/controller/PipelinesApiController.java#L46) TCL's SamUserFactory to check that the user is registered in Sam.

However, if a user is not registered in Sam/Terra, TCL is currently throwing an InternalServerErrorException that results in a 500 response, with the error message stating that the user was not found in Sam and an error code of 403 (FORBIDDEN). That's because TCL expects Sam to return a 404 (NOT_FOUND), so it doesn't properly handle the 403 [that Sam returns in this case ](https://github.com/broadinstitute/sam/blob/develop/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/StandardSamUserDirectives.scala#L241).

This PR updates the logic so that TCL treats both a 404 and a 403 as "User not found".

Teaspoons Jira ticket: https://broadworkbench.atlassian.net/browse/TSPS-404

### DIFF
--- a/src/main/java/bio/terra/common/iam/SamUserFactory.java
+++ b/src/main/java/bio/terra/common/iam/SamUserFactory.java
@@ -64,7 +64,7 @@ public class SamUserFactory {
     } catch (final NullPointerException e) {
       throw new UnauthorizedException(e.getMessage(), e);
     } catch (final ApiException e) {
-      if (e.getCode() == HttpStatus.NOT_FOUND.value()) {
+      if (e.getCode() == HttpStatus.NOT_FOUND.value() || e.getCode() == HttpStatus.FORBIDDEN.value()) {
         throw new UnauthorizedException("User not found", e);
       } else {
         throw new InternalServerErrorException(e);
--- a/src/main/java/bio/terra/common/iam/SamUserFactory.java
+++ b/src/main/java/bio/terra/common/iam/SamUserFactory.java
@@ -64,7 +64,8 @@ public class SamUserFactory {
     } catch (final NullPointerException e) {
       throw new UnauthorizedException(e.getMessage(), e);
     } catch (final ApiException e) {
-      if (e.getCode() == HttpStatus.NOT_FOUND.value() || e.getCode() == HttpStatus.FORBIDDEN.value()) {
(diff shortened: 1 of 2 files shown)


dvoet (approved):

pshapiro4broad (approved):
This seems correct with respect to Sam's behavior, but I wonder why 403/forbidden is being returned instead of 404/not found in the first place? I would expect a 404 if a user isn't found.

src/main/java/bio/terra/common/iam/SamUserFactory.java
```diff
@@ -64,7 +64,8 @@ public SamUser from(BearerToken bearerToken, String samBasePath) {
     } catch (final NullPointerException e) {
       throw new UnauthorizedException(e.getMessage(), e);
     } catch (final ApiException e) {
-      if (e.getCode() == HttpStatus.NOT_FOUND.value()) {
+      if (e.getCode() == HttpStatus.NOT_FOUND.value()
+          || e.getCode() == HttpStatus.FORBIDDEN.value()) {
         throw new UnauthorizedException("User not found", e);
       } else {
         throw new InternalServerErrorException(e);
```
pshapiro4broad, line 71: It seems odd to have an exception that contains a http status and then use it to map to a different http status. Another approach would be to map the ApiException's status directly to the new exception value
```
      throw new ErrorReportException(e, HttpStatus.valueOf(e.getCode())) {};
```

mmorgantaylor:
after some discussion offline, I'll move ahead with this PR. there are reasons Sam returns a 403 and reasons that this is remapped to a 401 that are beyond the scope of this PR to reconsider.

src/main/java/bio/terra/common/iam/SamUserFactory.java
mmorgantaylor: it's a fair point, but according to Core Services team there are reasons for returning a 401 here, and revisiting that is beyond the scope of this PR, so I'm not going to go there for now.

PR #228, opened 2025-03-03
CORE-69: bump k8s client-java lib

davidangb:
https://broadworkbench.atlassian.net/browse/CORE-69

supersedes #226

terra-common-lib had used the non-legacy versions of k8s client-java up until #220. This PR bumps client-java from the 22.x version to 23.x _and_ switches back to the non-legacy version.

The 23.x -legacy version shows compile problems in #226.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,7 +53,9 @@ dependencies {
     annotationProcessor group: 'org.springframework.boot', name: 'spring-boot-configuration-processor'
 
     // Misc. Services
-    implementation group: 'io.kubernetes', name: 'client-java', version: '22.0.0-legacy'
+    //  note that k8s client-java -legacy and non-legacy libraries are not interchangeable
+    //  see https://github.com/kubernetes-client/java?tab=readme-ov-file#release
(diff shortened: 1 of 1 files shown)


calypsomatic (approved):

kevinmarete (approved):

jsotobroad (approved):
thank you for digging into this!

LizBaldo:
@davidangb the change to use the legacy library was actually intentional as all of Leonardo relies on the old k8s java client structure and we do not have the bandwidth to refactor our code at the moment. 
What was the issue with the legacy version?

davidangb:
@LizBaldo Workspace Manager also uses TCL but has this comment:
https://github.com/DataBiosphere/terra-workspace-manager/blob/80766ecc3ad1a509e5b69ec4ad053ec46d7dcffd/service/dependencies.gradle#L79 … obviously we will have to coordinate something between the various TCL consumers!

LizBaldo:
Fun! Okay I will see what I can do to not use the legacy version moving forward

PR #248, opened 2025-09-03
[CORE-69] Update opentelemetry version

calypsomatic:
Updating `terra-common-lib` downstream is [causing](https://github.com/broadinstitute/sam/pull/1715) [failures](https://github.com/broadinstitute/rawls/pull/3469) (note that the first linked PR is currently passing due to reverting the TCL upgrade), apparently because in the imported version of `io.opentelemetry`, `ResourceAttributes` is deprecated.  Investigating the origin of opentelemetry in TCL revealed a note about version incompatibilities with spring boot < 3.5.  But TCL's spring boot version is not 3.5, so opentelemetry can now be updated.  This PR updates the specified opentelemetry version, but I'm not sure if that's the right way to go or if we actually want to NOT specify it since it should be pulled in through spring boot.  My attempt to remove the line in question resulted in failures, as did attempts to update to versions later than 2.14.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,12 +87,7 @@ dependencies {
     implementation group: 'ch.qos.logback.contrib', name: 'logback-jackson', version: '0.1.5'
 
     // OpenTelemetry dependencies:
-    // Spring Boot 3.4.3 still pulls in opentelemetry-bom 1.43.0.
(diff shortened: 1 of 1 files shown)


samanehsan (approved):

kevinmarete (approved):

fboulnois (approved):
👍

